PR #202, opened 2021-01-14
Switch to poetry-core

fabaff:
# Description

The build backend for Poetry was split out into a separate package `poetry-core`.


TomerFi:
@fabaff thank you.

Some of the workflow action doesn't work well with forks,
That's the reason for the test failure afaik.

I'll tend to this later, merging.

PR #484, opened 2022-05-30
feat: add switcher protocol type

thecode:
BREAKING CHANGE: SwitcherBridge broadcast ports changed to list

Signed-off-by: thecode <levyshay1@gmail.com>

# Pull Request

Add protocol types support to api and bridge

## Description
- Add protocol type to `DeviceType` enum
- Add `SwitcherType1Api` and `SwitcherType2Api` to api
- Fix `control_device` script to use new api
- Support bridge listening on multiple ports
- Fix `discover_devices` script to support multiple ports with user selection of ports or protocol type

## Checklist

## Additional information

> Anything else?
>

### DIFF
--- a/scripts/discover_devices.py
+++ b/scripts/discover_devices.py
@@ -24,26 +24,32 @@ from pprint import PrettyPrinter
 from aioswitcher.bridge import (
     SWITCHER_UDP_PORT_TYPE1,
     SWITCHER_UDP_PORT_TYPE2,
+    DeviceType,
     SwitcherBridge,
 )
 from aioswitcher.device import SwitcherBase
 
 printer = PrettyPrinter(indent=4)
 
-_examples = """Executing this script will print a serialized version of the discovered Switcher
+_examples = (
+    """Executing this script will print a serialized version of the discovered Switcher
 devices broadcasting on the local network for 60 seconds.
 You can change the delay by passing an int argument: discover_devices.py 30
 
 Switcher devices uses two protocol types:
-    Protocol type 1 (UDP port 20002), used by: V2, Touch, V4, Mini, Power Plug
-    Protocol type 2 (UDP port 20003), used by: Breeze, Runner, Runner Mini
+    Protocol type 1 (UDP port 20002), used by: """
+    + ", ".join(d.value for d in DeviceType if d.protocol_type == 1)
+    + """
+    Protocol type 2 (UDP port 20003), used by: """
+    + ", ".join(d.value for d in DeviceType if d.protocol_type == 2)
+    + """
 You can change the scanned protocol type by passing an int argument: discover_devices.py -t 1
 
 Note:
     WILL PRINT PRIVATE INFO SUCH AS DEVICE ID AND MAC.
 
 Example output:
-    Switcher devices broadcast a status message every approximantly 4 seconds. This
+    Switcher devices broadcast a status message every approximately 4 seconds. This
     script listens for these messages and prints a serialized version of the to the
     standard output, for example (note the ``device_id`` and ``mac_address`` properties)::
         {   'auto_shutdown': '03:00:00',
@@ -67,6 +73,7 @@ Print only protocol type 1 devices:
 Print only protocol type 2 devices:
     python discover_devices.py -t 2\n
 """  # noqa E501
+)
 
 parser = ArgumentParser(
     description="Discover and print info of Switcher devices",
--- a/scripts/discover_devices.py
+++ b/scripts/discover_devices.py
@@ -14,16 +14,35 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Python script for printing discovered Swithcer devices info.
+"""Python script for discovering Switcher devices."""
 
-Executing this script will print a serialized version of the discovered Switcher
-devices broadcasting on the local network for 60 seconds. You can change the delay
-by passing an int argument: ``discover_devices.py 30``.
+from argparse import ArgumentParser, RawDescriptionHelpFormatter
+from asyncio import get_event_loop, sleep
+from dataclasses import asdict
+from pprint import PrettyPrinter
+
+from aioswitcher.bridge import (
+    SWITCHER_UDP_PORT_TYPE1,
+    SWITCHER_UDP_PORT_TYPE2,
+    SwitcherBridge,
+)
+from aioswitcher.device import SwitcherBase
+
+printer = PrettyPrinter(indent=4)
+
+_examples = """Executing this script will print a serialized version of the discovered Switcher
+devices broadcasting on the local network for 60 seconds.
+You can change the delay by passing an int argument: discover_devices.py 30
+
+Switcher devices uses two protocol types:
+    Protocol type 1 (UDP port 20002), used by: V2, Touch, V4, Mini, Power Plug
+    Protocol type 2 (UDP port 20003), used by: Breeze, Runner, Runner Mini
+You can change the scanned protocol type by passing an int argument: discover_devices.py -t 1
 
 Note:
     WILL PRINT PRIVATE INFO SUCH AS DEVICE ID AND MAC.
 
-Example:
+Example output:
     Switcher devices broadcast a status message every approximantly 4 seconds. This
     script listens for these messages and prints a serialized version of the to the
     standard output, for example (note the ``device_id`` and ``mac_address`` properties)::
@@ -39,23 +58,21 @@ Example:
             'power_consumption': 0,
             'remaining_time': '00:00:00'}
 
-"""  # noqa: E501
+Print all protocol types devices for 30 seconds:
+    python discover_devices.py 30 -t all\n
 
-from argparse import ArgumentParser
-from asyncio import get_event_loop, sleep
-from dataclasses import asdict
-from pprint import PrettyPrinter
+Print only protocol type 1 devices:
+    python discover_devices.py -t 1\n
 
-from aioswitcher.bridge import (
-    SWITCHER_UDP_PORT_TYPE1,
-    SWITCHER_UDP_PORT_TYPE2,
-    SwitcherBridge,
+Print only protocol type 2 devices:
+    python discover_devices.py -t 2\n
+"""  # noqa E501
+
+parser = ArgumentParser(
+    description="Discover and print info of Switcher devices",
+    epilog=_examples,
+    formatter_class=RawDescriptionHelpFormatter,
 )
-from aioswitcher.device import SwitcherBase
-
-printer = PrettyPrinter(indent=4)
-
-parser = ArgumentParser(description="Discover and print info of Switcher devices")
 parser.add_argument(
     "delay",
     help="number of seconds to run, defaults to 60",
@@ -63,22 +80,14 @@ parser.add_argument(
     nargs="?",
     default=60,
 )
+possible_types = ["1", "2", "all"]
 parser.add_argument(
     "-t",
     "--type",
     required=False,
-    choices={1, 2},
-    help="set protocol type: 1 or 2",
-    type=int,
-)
-parser.add_argument(
-    "-p",
-    "--ports",
-    required=False,
-    help="bridge ports to listen on",
-    nargs="+",
-    type=int,
-    default=[SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2],
+    choices=possible_types,
+    help=f"set protocol type: {possible_types}",
+    type=str,
 )
 
 
@@ -97,14 +106,13 @@ async def print_devices(delay: int, ports: list[int]) -> None:
 if __name__ == "__main__":
     args = parser.parse_args()
 
-    if args.type == 1:
+    if args.type == "1":
         ports = [SWITCHER_UDP_PORT_TYPE1]
-    elif args.type == 2:
+    elif args.type == "2":
         ports = [SWITCHER_UDP_PORT_TYPE2]
     else:
-        ports = args.ports
+        ports = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2]
 
-    print(ports)
     try:
         get_event_loop().run_until_complete(print_devices(args.delay, ports))
     except KeyboardInterrupt:
--- a/scripts/discover_devices.py
+++ b/scripts/discover_devices.py
@@ -46,7 +46,11 @@ from asyncio import get_event_loop, sleep
 from dataclasses import asdict
 from pprint import PrettyPrinter
 
-from aioswitcher.bridge import SwitcherBridge
+from aioswitcher.bridge import (
+    SWITCHER_UDP_PORT_TYPE1,
+    SWITCHER_UDP_PORT_TYPE2,
+    SwitcherBridge,
+)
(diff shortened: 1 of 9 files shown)


src/aioswitcher/bridge.py
```diff
@@ -45,14 +45,16 @@
 logger = getLogger(__name__)
 
 
-SWITCHER_UDP_PORT = 20002
-SWITCHER_UDP_PORT2 = 20003
+# Type 1 devices: Heaters (v2, touch, v4), Plug
+SWITCHER_UDP_PORT_TYPE1 = 20002
+# Type 2 devices: Heaters Breeze, Runners
```
jafar-atili: I thought only Breeze and Runners broadcasting on 2003

thecode: Correct, this is a typo in the description, this is the real mapping for type 1 or 2:
https://github.com/TomerFi/aioswitcher/blob/8215236a5549dc44c4cbd956040c22e31911f79d/src/aioswitcher/device/__init__.py#L38-L46

scripts/discover_devices.py
```diff
@@ -59,24 +63,49 @@
     nargs="?",
     default=60,
 )
+parser.add_argument(
+    "-t",
+    "--type",
+    required=False,
+    choices={1, 2},
+    help="set protocol type: 1 or 2",
```
jafar-atili: I would list here the the DeviceType.value of each type/category
"set protocol type: 
        1 [Switcher Mini, Switcher Power Plug, etc..] 
        2[Breeze, Runner, Runner Mini]

jafar-atili: and also I think --type somehow conflict with --ports
I think we should have one.

I think we should have single option to control the bridge socket:
--device-category [cat1, cat2, all]

scripts/discover_devices.py
```diff
@@ -59,24 +63,49 @@
...
     args = parser.parse_args()
 
+    if args.type == 1:
+        ports = [SWITCHER_UDP_PORT_TYPE1]
+    elif args.type == 2:
+        ports = [SWITCHER_UDP_PORT_TYPE2]
+    else:
+        ports = args.ports
+
+    print(ports)
```
jafar-atili: leftover?

jafar-atili (approved):

scripts/discover_devices.py
thecode: agree, so I will remove the option to configure the ports, I also thought at start to call it device category but it conflicts with the category we already have in the code. So maybe we keep it as `protocol type`.

I will remove the option to set the port and keep only the option for protocol type, e.g.:
`type1, type2, all`

And add text which explains which devices use each type (as you wrote above)

scripts/discover_devices.py
```diff
@@ -14,16 +14,35 @@
...
+from aioswitcher.device import SwitcherBase
+
+printer = PrettyPrinter(indent=4)
+
+_examples = """Executing this script will print a serialized version of the discovered Switcher
+devices broadcasting on the local network for 60 seconds.
+You can change the delay by passing an int argument: discover_devices.py 30
+
+Switcher devices uses two protocol types:
+    Protocol type 1 (UDP port 20002), used by: V2, Touch, V4, Mini, Power Plug
```
jafar-atili:
```suggestion
    Protocol type 1 (UDP port 20002), used by: {", ".join([d.value for d in DeviceType if d.protocol_type == 1])}
```

scripts/discover_devices.py
```diff
@@ -14,16 +14,35 @@
...
+    Protocol type 2 (UDP port 20003), used by: Breeze, Runner, Runner Mini
```
jafar-atili:
```suggestion
    Protocol type 2 (UDP port 20003), used by: {", ".join([d.value for d in DeviceType if d.protocol_type == 2])}
```

scripts/discover_devices.py
jafar-atili: We already have this data, better than write it hardcoded

scripts/discover_devices.py
```diff
@@ -14,16 +14,35 @@
...
+from aioswitcher.bridge import (
+    SWITCHER_UDP_PORT_TYPE1,
+    SWITCHER_UDP_PORT_TYPE2,
+    SwitcherBridge,
+)
+from aioswitcher.device import SwitcherBase
+
+printer = PrettyPrinter(indent=4)
+
+_examples = """Executing this script will print a serialized version of the discovered Switcher
```
jafar-atili:
```suggestion
_examples = f"""Executing this script will print a serialized version of the discovered Switcher
```

jafar-atili (approved):

src/aioswitcher/device/__init__.py
```diff
@@ -35,23 +35,24 @@ class DeviceCategory(Enum):
...
+    V2_ESP = "Switcher V2 (esp)", "01a7", 1, DeviceCategory.WATER_HEATER
+    V2_QCA = "Switcher V2 (qualcomm)", "01a1", 1, DeviceCategory.WATER_HEATER
+    V4 = "Switcher V4", "0317", 1, DeviceCategory.WATER_HEATER
+    BREEZE = "Switcher Breeze", "0e01", 2, DeviceCategory.THERMOSTAT
+    RUNNER = "Switcher Runner", "0c01", 2, DeviceCategory.SHUTTER
+    RUNNER_MINI = "Switcher Runner Mini", "0c02", 2, DeviceCategory.SHUTTER
 
     def __new__(
-        cls, value: str, hex_rep: str, category: DeviceCategory
+        cls, value: str, hex_rep: str, protocol_type: int, category: DeviceCategory
```
TomerFi, line 49: what do protocol_type stands for?
can we have an enum instead?

src/aioswitcher/api/__init__.py
```diff
@@ -57,15 +57,17 @@
 
 logger = getLogger(__name__)
 
-SWITCHER_TCP_PORT2 = 10000
-SWITCHER_TCP_PORT = 9957
+# Type 1 devices: Heaters (v2, touch, v4), Plug
+SWITCHER_TCP_PORT_TYPE1 = 9957
```
TomerFi, line 61: If these are static ports, why not include them in the var name?
```python
SWITHCER_TCP_PORT_9957 = 9957
SWITCHER_TCP_PORT_10000 = 10000
```
?

src/aioswitcher/api/__init__.py
```diff
@@ -562,6 +563,34 @@ async def stop(self) -> SwitcherBaseResponse:
         return SwitcherBaseResponse(response)
 
 
+@final
```
TomerFi, line 566: Wouldn't it be better to have just one API class that takes the TCP port in its constructor?
Even and enum would be good I believe as a port identifier.

And if not, can we think of a title to better categorize _plugs & boilers_ vs. _thrmostats & runners_ instead of _type1_ vs. _type2_?

This just sounds too weird, if it's mandatory maybe consider boiler-plating this by creating similar classes to identify various types?

src/aioswitcher/api/packets.py
```diff
@@ -20,51 +20,51 @@
 NO_TIMER_REQUESTED = "00000000"
 NON_RECURRING_SCHEDULE = "00"
 # format values are local session id, timestamp
-REQUEST_FORMAT = "{}340001000000000000000000{}00000000000000000000f0fe"
-REQUEST_FORMAT2 = "{}390001000000000000000000{}00000000000000000000f0fe"
+REQUEST_FORMAT_TYPE1 = "{}340001000000000000000000{}00000000000000000000f0fe"
```
TomerFi, line 23: This still bothers me, we talked about it in the previous PRs, yet we merged it as-is.
Can we do better the TYPE1 vs TYPE2 for all the constants in this file?

If not, maybe consider a dynamic constants script injected by the initiator...

src/aioswitcher/bridge.py
```diff
@@ -268,14 +278,14 @@ def is_switcher_originator(self) -> bool:
             or len(self.message) == 159  # Switcher Runner and RunnerMini
         )
 
-    def get_ip(self) -> str:
-        """Extract the IP address from the broadcast message."""
+    def get_ip_type1(self) -> str:
```
TomerFi, line 281: going back to my common theme, can we do better than type1 vs type2?
we can even inject the get IP function based on the device type to make it more clear, no?

TomerFi (changes requested):

scripts/discover_devices.py
```diff
@@ -39,44 +64,63 @@
...
+    formatter_class=RawDescriptionHelpFormatter,
+)
 parser.add_argument(
     "delay",
     help="number of seconds to run, defaults to 60",
     type=int,
     nargs="?",
     default=60,
 )
+possible_types = ["1", "2", "all"]
```
TomerFi, line 90: the type1/2 is frustrating...
a better categorizing title would be much better.

TomerFi (changes requested):

tests/test_bridge.py
```diff
@@ -42,21 +42,21 @@ def udp_broadcast_server():
...
     assert_that(bridge.is_running).is_false()
     await bridge.stop()
-    mock_info.assert_called_with("udp bridge not started")
+    mock_info.assert_called_with("udp bridge on port %s not started", 1234)
 
 
 async def test_bridge_operation_as_a_context_manager(unused_udp_port_factory, mock_callback):
     port = unused_udp_port_factory()
-    async with SwitcherBridge(mock_callback, port) as bridge:
+    async with SwitcherBridge(mock_callback, [port]) as bridge:
```
TomerFi, line 53: Shouldn't we **also** use multiple ports to make this and other tests in this file better reflect the modification in this PR?
I think we should add tests that sort, or perhaps a better idea will be to parameterize the current test cases...

TomerFi (changes requested):

tests/test_bridge.py
thecode: I agree, tests should be added, this why I was aiming that we work on a branch so we can split this into multiple PRs, so a PR that add tests coverage should be made before we make the final merge to main.

src/aioswitcher/device/__init__.py
thecode: Yes, can you suggest something that will reflect the two protocol types? also note that we later had classes that are named Type1 and Type2

src/aioswitcher/api/__init__.py
thecode: Because the Type1 and Type2 aligns with the two protocol types in all other places. When you name it `SWITCHER_TCP_PORT_10000` it does not explain which devices or protocols related to this port. it only explains that it uses port 10000 which is irrelevant because if it was the only usage we could just eliminate this constant and set the port directly.

src/aioswitcher/api/packets.py
thecode: I am sorry I am not following you, we had a discussion that we agreed that we split all places that have to TYPE1 & TYPE2 so that it is clear that there are two types of protocols. As far as I see it if you don't agree with this base assumption (I see it repeating in comments). We should close this PR, discuss again and make a new one.

src/aioswitcher/bridge.py
thecode: Same as above....

scripts/discover_devices.py
thecode: Same...

src/aioswitcher/api/__init__.py
TomerFi: Yeah but the port itself is just an integer and is not directly related to the type of device using it as its broadcast port.
For future contributors, or even for our future self, "SWITCHER_TCP_PORT2" doesn't have too much meaning to it if you're not aware and/or make the connection that this correlated with the TYPE2 device.

src/aioswitcher/device/__init__.py
TomerFi: No, I actually can't think of anything...

src/aioswitcher/api/packets.py
TomerFi: I remember discussing it, I remember going along with this.
But I also remember constantly raising my concerns in regards to the naming issues with the new contributions,
and I remember getting constantly reassured that this is a WIP branch and we can address my naming concerns later.

I'll reiterate my concerns:
to anyone who's not involved with this upcoming release,
stuff like TYPE1/2, FORMAT1/2, and PORT1/2 don't mean much.

I understand the TYPE types and variables are a bit tricky to handle, AFAIK every type represents a set of devices that seemingly doesn't have a specific shared property that causes them to be in a specific TYPE group.
But I think that at the very least the PORT and PACKET_FORMAT variables can be named better.
Although, I would prefer the TYPEs to have even more meaning as well.

tests/test_bridge.py
TomerFi: Working on a WIP branch and "adding tests later" is ok, because it's relatively easy to locate missing tests.
However, realizing you need to modify an existing working test by just adding another test case to make it cover another end case,
I believe this is more easily noted before merging to the WIP branch unless, of course, you already took note of this and plan to include it when adding the new tests.
:-)

src/aioswitcher/api/packets.py
thecode: Can you suggest something with better meaning? The easiest way would be to review the changes and add code suggestions so I can commit them if they look good.
I don't have any better approach I can think of than what implemented here.

tests/test_bridge.py
thecode: I don't need to modify tests, the failing test is due to code coverage which is already missing on the base branch not on this PR. I can't do anything here until the tests PR is merged and rebasing this PR. This PR does not have any effect on coverage.
image

TomerFi: Note my initial https://github.com/TomerFi/aioswitcher/pull/484#discussion_r897784440 :
> Shouldn't we **also** use multiple ports

meaning, adding another test case regardless of failing/passing tests.

src/aioswitcher/api/packets.py
TomerFi: > Can you suggest something with better meaning? The easiest way would be to review the changes and add code suggestions so I can commit them if they look good.

So... you don't take reviews without alternative suggestions...
For me to be able to intervene further on a expected level, besides the suggestion I had already made in regards to the port variables naming, which you opposed, means I would have to get more invested in the next release, much like you and Jafar already are, and unfortunaly, I don't think I can spare the time in foreseen future.

You could have specified why is this such an issue and perhaps we could have brainstormed together,
or you could have stated that you've considered everything else but that's the only way and you own it (you are a maintainer here no?), instead, you took the if-you-don't-like-it-then-rewrite-it approach.

Dude, as I told you previously...
you are a maintainer here,
you maintain the main client of this repo (ha),
you are more involved with the users than I am and therefore know what is requested,
you are (partially) connected to the vendor,
you were involved in the planning of these modifications with Jafar.
I don't have the time to catch up on everything **you already know** in regards to this project,
so from my end, **whatever you say goes** - you have my blessing and I would try to help out whichever way I can.

To this, you replied that I can help out with code reviews,
yet it looks like you don't really like my code reviews.
:man_shrugging:

So...
Let's just keep this as-is,
I'll approve and merge this now.

PR #492, opened 2022-05-31
Handle special swing

jafar-atili:
# Pull Request

> First PR? don't hesitate, push it.

## Description

**Related issue (if any):** fixes #*issue_number_goes_here*

## Checklist

## Additional information

> Anything else?
>


jafar-atili:
Few comments about this PR:

1. It doesn't have to make it to the next release, it only means that the following AC remotes won't have vertical swing feature: `["ELEC7022", "ZM079055",  "ZM079065", "ZM079049", "ZM079065"]`

2. even with this PR there's a problem, the following remote types `["ELEC7022","ZM079055","ZM079065"]` have the same code, hence, their swing control is toggle which means that `control_device.py -s on -w on` can toggle the swing back to off.

3. Switcher have a problem reading the UDP/TCP swing state of ACs from item 1, they will always report that the swing is STOPped even if its working, so the this another problem that can make the UI swing state be out of sync with the actual swing state.

My 2 cents,
Jafar.

scripts/control_device.py
```diff
@@ -213,7 +213,7 @@
     choices=possible_states.keys(),
     required=True,
     help=f"thermostat state, possible values: {possible_states}",
-    default=None,
+    default="",
```
TomerFi: why do we need a default value for a required argument?

scripts/control_device.py
```diff
@@ -222,7 +222,7 @@
     choices=possible_modes.keys(),
     required=False,
     help=f"thermostat mode, possible values: {possible_modes}",
-    default=None,
+    default="",
```
TomerFi: based on the enum this possible values set is constructed from,
an empty string is not a valid option, why do we have it as a default value?

scripts/control_device.py
```diff
@@ -231,7 +231,7 @@
     choices=possible_fan_level.keys(),
     required=False,
     help=f"thermostat fan level, possible values: {possible_fan_level}",
-    default=None,
+    default="",
```
TomerFi: based on the enum this possible values set is constructed from,
an empty string is not a valid option, why do we have it as a default value?

scripts/control_device.py
```diff
@@ -240,15 +240,15 @@
     choices=possible_swing.keys(),
     required=False,
     help=f"thermostat swing, possible values: {possible_swing}",
-    default=None,
+    default="",
```
TomerFi: based on the enum this possible values set is constructed from,
an empty string is not a valid option, why do we have it as a default value?

scripts/control_device.py
```diff
@@ -240,15 +240,15 @@
...
 )
 control_thermostat_parser.add_argument(
     "-t",
     "--temperature",
     type=int,
     required=False,
     help=f"thermostat temperature, possible values: {possible_swing}",
-    default=None,
+    default=0,
```
TomerFi: temperature defaulting to 0?
that's a bit extreme, isn't it? do most thermostats even support this?

scripts/control_device.py
```diff
@@ -296,13 +296,39 @@ async def control_thermostat(
...
         printer.pprint(
             asdict(
                 await api.control_breeze_device(command),
                 verbose,
             )
         )
 
+        # Here is a special case to handle swing on special remotes
+        # if we entered here with only swing mode change, we perform it
+        if (
```
TomerFi: this `if statement` seems complex, can we simplify it?

src/aioswitcher/api/__init__.py
```diff
@@ -652,6 +652,13 @@ def __init__(self, ir_set: dict) -> None:
             }
         }
         """
+        self._separated_swing_command = self._remote_id in [
```
TomerFi: can you explain this list of specific devices IDs, please?

src/aioswitcher/api/__init__.py
```diff
@@ -704,6 +716,32 @@ def _lookup_key_in_irset(self, key: list):
                 # found a match, with modified list
                 return
 
+    def get_swing_command(self, swing: ThermostatSwing):
```
TomerFi, line 761: I think some tests might help clarify this function to those of us who don't know the internals of the thermostat swing thing, like me :-)

TomerFi (changes requested):

TomerFi:
This needs rebasing.

src/aioswitcher/api/__init__.py
jafar-atili: The developer from Switcher.
These are the only Air Conditioner remote IDs that have a special handling with the way you control their Swing capability

src/aioswitcher/api/__init__.py
jafar-atili: It is covered in the test multiple times as this function is major entry point for controlling Breeze

scripts/control_device.py
jafar-atili: because it is required for the control_thermostat function.

The argparse module won't assign value for this argument if its not required and no default value specified

scripts/control_device.py
jafar-atili: because it is required for the control_thermostat function.

The argparse module won't assign value for this argument if its not required and no default value specified

scripts/control_device.py
jafar-atili: because it is required for the control_thermostat function.

The argparse module won't assign value for this argument if its not required and no default value specified

scripts/control_device.py
jafar-atili: because it is required for the control_thermostat function.

The argparse module won't assign value for this argument if its not required and no default value specified

scripts/control_device.py
jafar-atili: its the default value of int the same way the default value for string is the empty string.

in the control_thermostat we ignore zero value.

and if a lower number than the minimum temperature is provided we set the minimum temperature supported by the AC, same thing for maximum temperature

scripts/control_device.py
jafar-atili: I think its well explained in the comment below it, wasn't able to simplify it more than this

scripts/control_device.py
TomerFi: my question was why do we need a default value (based on it being required), not why is it required...

a "required" marked argument is an argument that is required as input by the user,
the argparse module will break the parsing process if it's missing.

an argument that has a "default" value is, by logic, not a "required" argument,
if the user doesn't input it, the argparse module uses the default value for it.

using both "required" and "default" makes no sense, please remove one.
unless, of course, I'm missing something, if so, please try explaining why again.

scripts/control_device.py
TomerFi: - how can an empty string be a default value as an enum key conversation?
- the "required=false" is the default value afaik, remove this property.

scripts/control_device.py
TomerFi: see above https://github.com/TomerFi/aioswitcher/pull/492#discussion_r912333604

scripts/control_device.py
TomerFi: see above https://github.com/TomerFi/aioswitcher/pull/492#discussion_r912333644

scripts/control_device.py
TomerFi: see above https://github.com/TomerFi/aioswitcher/pull/492#discussion_r912333644

scripts/control_device.py
TomerFi: see above https://github.com/TomerFi/aioswitcher/pull/492#discussion_r912333644

scripts/control_device.py
TomerFi: > its the default value of int the same way the default value for string is the empty string.

you don't say... :wink:
note that for the above properties you are referencing with this note,
you shouldn't have used an empty string, to begin with. :upside_down_face:

> in the control_thermostat we ignore zero value.

why 0? have you ever seen an AC unit with a temperature lower than 16 (15 to be safe)?
(maybe I should have asked this in the previous PRs)

> and if a lower number than the minimum temperature is provided we set the minimum temperature supported by the AC, same thing for maximum temperature

given https://github.com/TomerFi/aioswitcher/pull/492#discussion_r912333604
why not just remove the "default" property, and set the "required" one to true?
we can even go the extra mile and note the "zero/<15 = max conversation" in the help property.

src/aioswitcher/api/__init__.py
TomerFi: > It is covered in the test multiple times as this function is major entry point for controlling Breeze

can you provide a reference?

src/aioswitcher/api/__init__.py
TomerFi: please set this list as a "constant" and comment better about what it does and the actual ac types it refers to.

scripts/control_device.py
TomerFi: no, it actually doesn't.
the comment may explain why is this statement there,
but it doesn't make it less complex to maintain/debug.

jafar-atili:
@TomerFi, your Git abilities is required here, I tried to rebase on the local repo, now it doesn't like some commit messages

TomerFi:
@jafar-atili 
I'm working on this - will be done soon.

TomerFi (approved):

PR #559, opened 2022-09-25
chore: switched codebase to strict typing

TomerFi:
- fix: typo in message parser function signature
- chore: adjutments for strict typing

# Pull Request

> First PR? don't hesitate, open it.

## Description

## Checklist

## Additional information

> Anything else?
>

### DIFF
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -22,7 +22,7 @@ from logging import getLogger
 from socket import AF_INET, inet_ntoa
 from struct import pack
 from types import TracebackType
-from typing import Any, Callable, Dict, Optional, Tuple, Type, final
+from typing import Any, Callable, Dict, List, Optional, Tuple, Type, final
 from warnings import warn
 
 from .device import (
@@ -173,13 +173,13 @@ class SwitcherBridge:
     def __init__(
         self,
         on_device: Callable[[SwitcherBase], Any],
-        broadcast_ports: list[int] = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2],
+        broadcast_ports: List[int] = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2],
     ) -> None:
         """Initialize the switcher bridge."""
         self._on_device = on_device
         self._broadcast_ports = broadcast_ports
         self._is_running = False
-        self._transports = {}  # type: Dict[int, Optional[BaseTransport]]
+        self._transports: Dict[int, Optional[BaseTransport]] = {}
 
     async def __aenter__(self) -> "SwitcherBridge":
         """Enter SwitcherBridge asynchronous context manager."""
@@ -237,14 +237,14 @@ class UdpClientProtocol(DatagramProtocol):
 
     def __init__(self, on_datagram: Callable[[bytes], None]) -> None:
         """Initialize the protocol."""
-        self.transport = None  # type: Optional[BaseTransport]
+        self.transport: Optional[BaseTransport] = None
         self._on_datagram = on_datagram
 
     def connection_made(self, transport: BaseTransport) -> None:
         """Call on connection established."""
         self.transport = transport
 
-    def datagram_received(self, data: bytes, addr: Tuple) -> None:
+    def datagram_received(self, data: bytes, addr: Tuple[Any, Any]) -> None:
         """Call on datagram received."""
         self._on_datagram(data)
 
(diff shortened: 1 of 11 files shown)


src/aioswitcher/bridge.py
```diff
@@ -173,13 +173,15 @@ class SwitcherBridge:
     def __init__(
         self,
         on_device: Callable[[SwitcherBase], Any],
-        broadcast_ports: list[int] = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2],
+        broadcast_ports: List[int] | None = None,
```
thecode: This changes the functionality of the code and will break existing implementation
```suggestion
        broadcast_ports: List[int] | None = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2],
```

TomerFi: @thecode good catch!
that was actually suggested by the _sonarlint_ extension,
I must have committed this by accident, my bad.

I think your suggestion was what _sonarlint_ complained about to begin with,
let me see if I can work something out that will both satisfy _sonarlint_ and not break anything.

That being said, the tests pass on my computer,
so I guess we don't test this potential breaking point very well.

(I don't think the CI failures are related to , the same tests pass on my computer.
so I guess it's missing requirement or something, but I haven't had the change to work on this yet).

EDIT:
The tests failed because the typing operator `|` was introduced in py3.10, fixed here https://github.com/TomerFi/aioswitcher/pull/559/commits/05db97bfc4c4d5859fde8606a9707f32d31cbffc.

TomerFi: image

TomerFi: @thecode 
fixed here https://github.com/TomerFi/aioswitcher/pull/559/commits/b5e7e76aeeb6791c71899979dccb71321cf67028, please take a look.

src/aioswitcher/bridge.py
```diff
@@ -173,13 +173,15 @@ class SwitcherBridge:
     def __init__(
         self,
         on_device: Callable[[SwitcherBase], Any],
-        broadcast_ports: list[int] = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2],
+        broadcast_ports: Union[List[int], None] = None,
     ) -> None:
         """Initialize the switcher bridge."""
+        if broadcast_ports is None:
+            broadcast_ports = [SWITCHER_UDP_PORT_TYPE1, SWITCHER_UDP_PORT_TYPE2]
```
thecode: Why not assigning them in line 176?

thecode:
As I wrote on Discord, I think we should make our typing Python 3.10 compatible and avoid need to change it in the future

TomerFi:
> As I wrote on Discord, I think we should make our typing Python 3.10 compatible and avoid need to change it in the future

and as I explained in Discord, we're still officially supporting 3.9...
and this project is fully 3.10 compatible, I think you are mistaken.

src/aioswitcher/bridge.py
TomerFi: https://github.com/TomerFi/aioswitcher/pull/559#discussion_r979384122

thecode: This is not related, you can still assign them in Python 3.9

TomerFi: it doesn't really matter...
assigning in row 180 or 176, as I explained above this was reported as a codesmell by sonarlint.

this might be true or might be a false-positive,
I could either investigate why is this a false-positive, change this code to adhere the warning, or disable this specific sonarlint rule.

I chose to make the change and not disable the rule because it makes absolutely not difference from the code perspective and I don't have any time to spare for investigating sonarlint's false-positives...

but... if you're worried by this small meaningless modification, I can reverse it and disable sonarlint...
your call.

src/aioswitcher/bridge.py
thecode: > > As I wrote on Discord, I think we should make our typing Python 3.10 compatible and avoid need to change it in the future
> 
> and as I explained in Discord, we're still officially supporting 3.9... and this project is fully 3.10 compatible, I think you are mistaken.

Check other projects and see that they align code with the most recent supported version and not the older one, they just check backward compatibility vs oldest supported version. It looks weird to make a PR to align code to older version.
However it is your repo, feel free to merge but IMO at least the comment about `broadcast_ports` should be fixed, it is not related to a Python version (I am doing it on projects that are still Python 3.7 compatible with typing.

thecode (approved):
Thanks!

PR #561, opened 2022-10-07
chore: Avoid get_event_loop() deprecation warning in discover_devices.py

oranja:
# Pull Request

## Description

From Python 3.11, it will not be valid to call `get_event_loop()` if there is no running event loop already.

Python 3.10 already shows deprecation warnings:
> DeprecationWarning: There is no current event loop

From the [docs](https://docs.python.org/3.10/library/asyncio-eventloop.html#asyncio.get_event_loop):
> Deprecated since version 3.10: Deprecation warning is emitted if there is no running event loop. In future Python releases, this function will be an alias of get_running_loop().

Instead, this commit switches to `asyncio.run(..)` to create an event loop and start the device scanning.

## Checklist

### DIFF
--- a/scripts/discover_devices.py
+++ b/scripts/discover_devices.py
@@ -16,8 +16,8 @@
 
 """Python script for discovering Switcher devices."""
 
+import asyncio
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
-from asyncio import get_event_loop, sleep
 from dataclasses import asdict
 from pprint import PrettyPrinter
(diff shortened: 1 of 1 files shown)


thecode (dismissed):

thecode (approved):
Thanks!

oranja:
Thank you for the quick review!
(sorry for the mess while sorting out the different commit checks. haven't done open source in a long time)

PR #567, opened 2022-10-11
build: introduced gnu make

TomerFi:
# Description

Incorporate [make](https://www.gnu.org/software/make/manual/make.html) for managing build tasks without breaking current usage of _poethepoet_.

**Related issue (if any):** resolves #564 

# Checklist

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -48,9 +48,9 @@ jobs:
 
       - name: Lint the project
         run: |
-          poetry install --no-interaction
-          poetry run poe lint
-          poetry run poe lic_check
+          make install-all-deps
+          make lint
+          make verify-license-headers
 
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/pr.yml
```diff
@@ -151,5 +152,5 @@ jobs:
 
       - name: Build documentation site
         run: |
-          poetry install --no-interaction
-          poetry run poe docs_build
+          make install-all-deps
+          make docs-build
```
oranja, line 156: Hi :)

1. I know this isn't new, but this commit made me wonder, why are all the deps required here, rather than just the docs deps?

2. If the docs deps are enough here, like in the other places, I wonder next if it wouldn't be even more helpful for local development to configure the `docs-build` target to always depend on `install-docs-only-deps`? (It's been ages, so I can't suggest the actual syntax off the top of my head)
It might be annoying to have the dependencies check on each local iteration, but sometimes the dependencies themselves change during development, and perhaps the overhead isn't that bad? idk, just thinking out loud...

TomerFi: Nice catch! 👏🏻 

This is intentionally though, the _pr_ workflow has 3 jobs: _lint_, _test_, _docs_.
The first two only needs the dev dependencies, the last one only needs the docs dependencies.
But if want to share the cache between all 3 jobs... 😁 

** the first job, _lint_ will install all dependencies, the next ones, _test_, and _docs_, won't install anything.

Our other workflows, _pages_, _release_, and _stage_, all have just one job, so we can be more specific with our dependency installation.

** the _release_ workflow, actually has 2 jobs, but the first one is just for collecting the required data and doesn't require a cache.

thecode (approved):

PR #642, opened 2023-04-06
feat: include bridge port 10002, used in new devices in broadcast discovery

### LINKED ISSUE #592: switcher version 5.3 is not working
### What happened?

A new release (5.3) of switchers are rolling out and the support of this library is lost

### Module Version

latest (HA)

### Device Type

Switcher Mini

### Firmware Version

5.3

### Relevant log output

@marcipar can you provide logs, please?

```shell
```

liadav:
## Description

> In newer firmware version of Switcher V2, the broadcast port changed from 20002 to 10002. I added this port to the bridge discovery code.

**Related issue (if any):** fixes #592

## Checklist

- [V] I have followed this repository's contributing guidelines.
- [V] I will adhere to the project's code of conduct.

### DIFF
--- a/docs/scripts.md
+++ b/docs/scripts.md
@@ -20,7 +20,7 @@ devices broadcasting on the local network for 60 seconds.
 You can change the delay by passing an int argument: discover_devices.py 30
 
 Switcher devices uses two protocol types:
-    Protocol type 1 (UDP port 20002), used by: Switcher Mini, Switcher Power Plug, Switcher Touch, Switcher V2 (esp), Switcher V2 (qualcomm), Switcher V4
+    Protocol type 1 (UDP port 20002 or port 10002), used by: Switcher Mini, Switcher Power Plug, Switcher Touch, Switcher V2 (esp), Switcher V2 (qualcomm), Switcher V4
     Protocol type 2 (UDP port 20003), used by: Switcher Breeze, Switcher Runner, Switcher Runner Mini
 You can change the scanned protocol type by passing an int argument: discover_devices.py -t 1
 
--- a/scripts/discover_devices.py
+++ b/scripts/discover_devices.py
@@ -38,7 +38,7 @@ devices broadcasting on the local network for 60 seconds.
 You can change the delay by passing an int argument: discover_devices.py 30
 
 Switcher devices uses two protocol types:
-    Protocol type 1 (UDP port 20002), used by: """
+    Protocol type 1 (UDP port 20002 or 10002), used by: """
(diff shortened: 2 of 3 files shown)


TomerFi:
Thank you @liadav.
This is a "hot topic", more people should be involved.

@dmatik Do you mind being involved in this too?
Do you know if we have a way to differentiate the new type1 from the old ones, so we can perhaps set the port in code?

dmatik:
Sure. I can help. 
Not sure about any way to differentiate. 
I have installed recently new Switcher Touch V3. 
Did not have issue to retrieve the ID using the old script. 
I am still using the WebAPI approach, not integration.

liadav:
Based on my initial research, it looks like the only thing that's different is the broadcast port. So, I think we just need to add that for the discovery phase. It shouldn't matter whether it's a new type1 or an old one.
By the way, if you want, we can run some tests on my device to make sure everything is working properly.

TomerFi:
> Sure. I can help.
> Not sure about any way to differentiate.
> I have installed recently new Switcher Touch V3.
> Did not have issue to retrieve the ID using the old script.
> I am still using the WebAPI approach, not integration.

@dmatik Oh yeah, you're right.
I was away from this project for so long, I forgot some of the implementation specifications.
After reviewing this, I now remember that the this port is used only by the Bridge, and not the API.
So we probably, don't need to do anything for the WebAPI as it doesn't use the Bridge.
:-)

TomerFi:
> Based on my initial research, it looks like the only thing that's different is the broadcast port. So, I think we just need to add that for the discovery phase. It shouldn't matter whether it's a new type1 or an old one. By the way, if you want, we can run some tests on my device to make sure everything is working properly.

@liadav Thank you very much for you contribution.
Do you mind having a look at the [discover_devices](https://github.com/TomerFi/aioswitcher/blob/dev/scripts/discover_devices.py) script? It also uses the same port variables.

We also need to include it in the script's help sections, which mentions the original ports only.

liadav:
Sure, I added both to [discover_devices](https://github.com/TomerFi/aioswitcher/blob/dev/scripts/discover_devices.py) script and the script's help section.

TomerFi:
LGTM
Thank you @liadav .

@dmatik I think we can merge this now, WDYT?

TomerFi (approved):

dmatik:
> LGTM
> 
> Thank you @liadav .
> 
> 
> 
> @dmatik I think we can merge this now, WDYT?

Yes, sure.

TomerFi:
@all-contributors add @liadav for code

PR #710, opened 2023-09-15
chore: stop shutter support for all known shutter devices

YogevBokobza:
## Description

> Added DeviceType to all devices.
This is preparation for the new switcher devices that will be added.
Part one of splitting this PR https://github.com/TomerFi/aioswitcher/pull/694

This code is needed because the current state is that devices are defined as categories such as Runner, Power Plug, etc.
But there are different logic needed per specific devices such as Runner VS Runner S11.
This code will be able to determine different device types within the same category

BREAKING CHANGE: API users are now required to pass their device type as an argument

**Related issue (if any):** https://github.com/TomerFi/aioswitcher/pull/694

## Checklist

- [x] I have followed this repository's contributing guidelines.
- [x] I will adhere to the project's code of conduct.

## Additional information

> Anything else?

PR #712, opened 2023-09-20
test: added breeze api test cases

YogevBokobza:
## Description

> Getting 100% code coverage by remove unused code and add tests

## Checklist

## Additional information

> Anything else?

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,5 @@
+[report]
+exclude_lines =
+    raise NotImplementedError
+    raise RuntimeError
+    except KeyError
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,0 @@
-[report]
-exclude_lines =
-    raise NotImplementedError
-    raise RuntimeError
-    except KeyError
--- a/tests/test_api_tcp_client.py
+++ b/tests/test_api_tcp_client.py
@@ -18,7 +18,6 @@ import os
 from asyncio.streams import StreamReader, StreamWriter
 from binascii import hexlify, unhexlify
 from datetime import timedelta
-from json import loads
 from unittest import skipUnless
 from unittest.mock import AsyncMock, Mock, patch
 
@@ -89,32 +88,6 @@ async def connected_api_type2(reader_mock, writer_mock):
         await api.connect()
         yield api
         await api.disconnect()
-
-
-class MockResponse:
-    """Mock response for the aiohttp response object."""
-
-    def __init__(self, text, status):
-        """Initialize the aiohttp response object."""
-        self._text = text
-        self.status = status
-        self.content_length = len(self._text)
-
-    async def text(self):
-        """Text data of the response class mock."""
-        return self._text
-
-    async def json(self, content_type=None):
-        """Json method of the response class mock."""
-        return loads(self._text)
-
-    async def __aexit__(self, exc_type, exc, tb):
-        """Magic method when starting an async block."""
-        pass
-
-    async def __aenter__(self):
-        """Magic method when leaving an async block."""
-        return self
 
 
 @patch("logging.Logger.info")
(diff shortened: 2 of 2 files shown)


tests/test_api_tcp_client.py
```diff
@@ -328,6 +308,13 @@ async def test_breeze_build_command_function_invalid_mode(resource_path_root):
         remote.build_command(DeviceState.ON, ThermostatMode.COOL, 20, ThermostatFanLevel.AUTO, ThermostatSwing.ON, DeviceState.OFF)
 
 
+async def test_breeze_build_command_function_specific_case():
+    remote_manager = SwitcherBreezeRemoteManager()
+    remote_7001 = remote_manager.get_remote("ELEC7001")
+    command = remote_7001.build_command(DeviceState.OFF, ThermostatMode.COOL, 20, ThermostatFanLevel.HIGH, ThermostatSwing.ON, DeviceState.ON)
+    assert_that(command.command).is_equal_to("00000000524337327c32317c33327c32367c34437c39387c537c32327c30337c373237325b32325d7c39383841303030303830")
```
TomerFi, line 315: It's been a long time since I worked on this project, I don't remember what goes into these packets. Are you there's isn't anything important here?

.coveragerc
```diff
@@ -0,0 +1,5 @@
+[report]
+exclude_lines =
+    raise NotImplementedError
+    raise RuntimeError
+    except KeyError
```
TomerFi: If this is here for the sole purpose of reaching 100% code coverage, we can drop it.
I'm ok with our current 98%. :-)

TomerFi (changes requested):

tests/test_api_tcp_client.py
YogevBokobza: Yes.
As part of this coverage improvement, I just saw that part of a code didn't reach by any test so I added a test to support that.

In this file, There is also part of the code which is not used anywhere..

.coveragerc
YogevBokobza: Why drop it?
Those exclusions make sense and improve coverage and also are sensitive to decreased coverage over-time (As did happen in [#710)](https://github.com/TomerFi/aioswitcher/pull/710)

TomerFi: 1) What's wrong with 98% coverage? Who says 100% is required or even a best practice?
2) If we were too add this, and I don't think we should even bother, we use `pyproject.toml` for coverage configuration.

tests/test_api_tcp_client.py
TomerFi: Yes, I noticed that - thanks.
I was only talking about the packet itself, the one you used in the `assert_that` statement, just making sure it doesn't contain any personal/sensitive data before merging.

YogevBokobza: No its not. its the same packet as used before in the code so it should be fine :)

.coveragerc
YogevBokobza: Well, I don't know then.. I think if it's possible to be at 100% why not.
Those excludes make sense in my opinion as why would you test abstract class? why would you test RuntimeError?

It's your call. 
If you want you can merge this PR if not then no..

TomerFi: Let's remove this file, seems redundant to me.

YogevBokobza: As you wish.. File removed.

TomerFi (approved):

.coveragerc
TomerFi: Thank you very much for your hard work - merging this now.
For future reference, this repository is already full of annoying configuration files, whenever possible, it's prefered using `pyproject.toml'. Coverage is already configured there:

https://github.com/TomerFi/aioswitcher/blob/0544a5047ae1e9ae738f5b909b67c9af793e355e/pyproject.toml#L100

TomerFi:
@allcontributors add @YogevBokobza for test and maintenance

PR #716, opened 2023-09-27
chore: added script and utility method for token retrieval

YogevBokobza:
## Description

> Following our discussion, Switcher made an API call to get the Token.
I made this little script and its logic including tests.
Preparation step for Token based devices.

## Checklist

## Additional information

> Anything else?


requirements.txt
```diff
@@ -1 +1,2 @@
 poetry==1.5.1
+types-requests==2.31.0.6
```
TomerFi: This goes in pyproject.toml, it's not required here.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ pytest-mockservers = "^0.6.0"
 pytest-sugar = "^0.9.4"
 time-machine = "^2.7.0"
 yamllint = "^1.26.3"
+types-requests = "^2.31.0.6"
```
TomerFi: This goes in `tool.poetry.dependencies` not the `dev` group.

src/aioswitcher/device/tools.py
```diff
@@ -158,3 +160,25 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+        try:
+            json_response = response.json()
+            print("JSON Response:", json_response)
+            token = json_response.get("token", "")
+
+        except ValueError:
+            print("Response content is not valid JSON")
+    else:
+        print("Request failed with status code:", response.status_code)
+    return token
```
TomerFi, line 187: Let's encapsulate the token with a value class, and return in as a JSON. It will be more robust and won't break if we'll need to modify something response's body.

scripts/get_token.py
```diff
@@ -0,0 +1,61 @@
...
+    type=str,
+)
+
+
+if __name__ == "__main__":
+    try:
+        args = parser.parse_args()
+
+        token = get_token(args.username, args.password)
+        print("Your Token is: " + token)
```
TomerFi, line 58: If we encapsulate the response with a value class, it will be probably best to print the returning JSON as-is. Maybe just prettify it or something.

TomerFi (changes requested):
Overall this looks good, but maybe we should wait a bit and see how the discussion goes.

EDIT:
Waiting is not mandatory. This doesn't affect the library itself, I'm ok with merging this after you address the requested changes.

EDIT2:
Having second thoughts here.
Let's remove the `get_token` utility function and place the entire logic in the `get_token` script. So far we managed to keep this library runtime-dependency free. The _scripts_ are a part of the project but not part of the library itself. If `requests` is only required for the script, we don't need it, nor do we need the utility function, in the library.

requirements.txt
YogevBokobza: Change has done :)

pyproject.toml
YogevBokobza: Change has done :)

src/aioswitcher/device/tools.py
YogevBokobza: Do you have an example of what you mean?

TomerFi: Sure, take a look at the classes extending the the base response in the [api module](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/api/messages.py#L146). Basically they encapsulate responses we got from the device.

TomerFi: But I'm actually having second thoughts here, after thinking about it over night.
I'll write my concerns in the PR comments.

YogevBokobza:
@TomerFi 
Regarding moving the get_token to the scripts only, my intention with this PR is that have the get_token logic inside the library and call it from HA and the WebAPI repo..
Moving this to the script means it won't be possible to get the token by code in the other repos and we are back to once again do a manual step..

TomerFi:
> @TomerFi Regarding moving the get_token to the scripts only, my intention with this PR is that have the get_token logic inside the library and call it from HA and the WebAPI repo.. Moving this to the script means it won't be possible to get the token by code in the other repos and we are back to once again do a manual step..

As I said in the discussion, I prefer it to remain a manual step - This library and its HA integration are currently running local-only - and I would like to keep it that way for as long as possible.
If we decide in the discussion to change that, than by all means... but this decision is not made yet.

src/aioswitcher/device/tools.py
YogevBokobza: Take a look now after using the value class.

scripts/get_token.py
YogevBokobza: Take a look now after using the value class.

src/aioswitcher/device/tools.py
```diff
@@ -158,3 +160,23 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+    if response.status_code == 200:
+        print("Request successful")
+        try:
+            token = DeviceToken(response.json())
+
+        except ValueError:
+            print("Response content is not valid JSON")
+    else:
+        print("Request failed with status code:", response.status_code)
+    return token
```
thecode: Please do not use `print` in API calls. You already have prints in the script intended for the user. If we plan to use this API for automating the token these prints should be replaced by debug logger.

tests/test_device_tools.py
```diff
@@ -116,3 +116,17 @@ def test_convert_str_to_devicetype_should_return_expected_devicetype(str, type):
...
+    ])
+def test_get_token_should_return_valid_token(username, password, token):
+    assert_that(tools.get_token(username, password).to_json()).is_equal_to(token)
+
+
+@mark.parametrize("username, password, token", [
+    ("username", "password", '{"error": "Invalid credentials"}')
+    ])
+def test_get_token_with_error_should_return_empty_token(username, password, token):
+    assert_that(tools.get_token(username, password).to_json()).is_equal_to(token)
```
thecode: I may be missing something, but does this test doing a real API call to Switcher servers? If it does, you should patch it not to do any externals IO/network calls.

YogevBokobza: Yes actually yes.. they made this user and should stay for ever..
How can we simulate the api call otherwise?
its the get token function we are testing.

src/aioswitcher/device/tools.py
YogevBokobza: Ok sure will do

YogevBokobza: Done. replace print with logger

tests/test_device_tools.py
YogevBokobza: Fixed to mock the request post itself.

src/aioswitcher/device/tools.py
TomerFi: I don't it's needed anymore if we use it only for the script. We don't need this functions in the module itself.

scripts/get_token.py
TomerFi: I don't it's needed anymore if we use it only for the script. We don't need this functions in the module itself.

src/aioswitcher/device/tools.py
YogevBokobza: No actually. Shay and I on the same point here.
This get token logic should be part of this library in order to be able to call it from HA and the other repo code.

scripts/get_token.py
YogevBokobza: No actually. Shay and I on the same point here.
This get token logic should be part of this library in order to be able to call it from HA and the other repo code.

src/aioswitcher/device/tools.py
TomerFi: Are we going to use this function from the HA?

scripts/get_token.py
TomerFi: Are we going to use this function from the HA?

src/aioswitcher/device/tools.py
YogevBokobza: Here. Yes.

scripts/get_token.py
YogevBokobza: This is not for HA but for using this library only.
The internal function is for HA

TomerFi: You said "Shay and I on the same point here" so I assumed you are talking about HA.

As I said in my previous comments, I don't think it's needed inside the library, it's not worth adding `requests` as a dependency to a project that has no dependencies. Users can use the script or the vendor's site for getting this token, and I'm ok with users getting their token by themselves.

TomerFi: That's unless, of course, HA integration requires it.

YogevBokobza: This ia what Shay said:
"Since Switcher already added support for an API, I would prefer it will be a part of the library, It can be part of the tools as suggested in https://github.com/TomerFi/aioswitcher/pull/716 so that it doesn't run on runtime unless requested by the user (or HA integration). Library itself would still stay local only, but will provide a util to get the token."

TomerFi: I'm not following.
You said: "This is not for HA but for using this library only."

Will the the `get_token` function will be used by HA integration or not?
If not, and it will only be used by the script, why do we need it in the library? why not in the script?

YogevBokobza: I will try to simpify this:
get_token function in src/aioswitcher/device/tools.py will be used in HA.
In HA we will call this function to get the token by code.
This scripts/get_token.py is for people who wants to use or test this library as it is.

TomerFi: @thecode are we ok with having the integration use the vendor's api?

TomerFi (approved):

TomerFi (approved):

PR #724, opened 2024-01-04
fix: device stop working because lacking of device_key

### LINKED ISSUE #725: switcher touch is detected but commands not working
### What happened?

switcher touch (hot water controller - boiler) is discovered with discover_devices.py -t 1 but all attempt to run commands against it fail (attempted, on / off / get_status).

### Module Version

dev branch

### Device Type

Switcher Touch (V3)

### Firmware Version

5.15

### Relevant log output

```shell
running python discover_devices.py -t 1  

{   'auto_shutdown': '01:30:00',
    'device_id': 'xxxxxx',
    'device_state': <DeviceState.OFF: ('00', 'off')>,
    'device_type': <DeviceType.TOUCH: ('Switcher Touch', '030b', 1, <DeviceCategory.WATER_HEATER: 1>)>,
    'electric_current': 0.0,
    'ip_address': 'xxx.xxx.xx.xx',
    'last_data_update': datetime.datetime(2024, 1, 5, 9, 37, 53, 651814),
    'mac_address': 'xx:xx:xx:xx:xx:xx',
    'name': 'xxxx',
    'power_consumption': 0,
    'remaining_time': '00:00:00'
}

but all attempts to run commands against it fail:
e.g:
python control_device.py get_state -d xxxx -i xxx.xxx.xx.xx

Traceback (most recent call last):
  File "/Users/xxxx/PycharmProjects/aioswitcher/src/aioswitcher/api/__init__.py", line 369, in get_state
    response = SwitcherStateResponse(state_resp)
  File "<string>", line 4, in __init__
  File "/Users/xxxx/PycharmProjects/aioswitcher/src/aioswitcher/api/messages.py", line 200, in __post_init__
    self.state = parser.get_state()
  File "/Users/xxxxx/PycharmProjects/aioswitcher/src/aioswitcher/api/messages.py", line 85, in get_state
    return states[hex_state]
KeyError: ''

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/xxxx/PycharmProjects/aioswitcher/scripts/control_device.py", line 429, in <module>
    asyncio.run(get_state(args.device_id, args.ip_address, args.verbose))
  File "/Users/xxxx/anaconda3/envs/electra/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/xxxx/anaconda3/envs/electra/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/Users/xxxx/PycharmProjects/aioswitcher/scripts/control_device.py", line 298, in get_state
    printer.pprint(asdict(await api.get_state(), verbose))
  File "/Users/xxxx/PycharmProjects/aioswitcher/src/aioswitcher/api/__init__.py", line 373, in get_state
    raise RuntimeError("get state request was not successful") from ve
RuntimeError: get state request was not successful

after doing some debugging what i see is that after self._login() function pass (it does pass) there is attempt to send get_state package but the response is b'' (empty)

(worth mentioning that the hot water controll, does work with switchers app well so there is no problem with the device)
```

YogevBokobza:
## Description

> Fix provided by Switcher to solve this issue: when devices get shut down (For example Switcher Touch device) and restored you can no longer control it. This is shown in the Home Assistant UI.
In this case, only using the "Switching to Local" method helps.
This will solve this all together.

The fix:
I am retrieving device_key from the UDP message which is the unique device login key used for the login1 method.
This is mainly for Boiler devices and works in all versions.

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@thecode  @TomerFi  Can you review please?.
This is ready to go.

YogevBokobza:
Added a script to get device_key for a device. (only for using this library as it is)
In HA integration we will take this variable from the discovery.

scripts/control_device.py
```diff
@@ -75,6 +75,13 @@
     required=True,
     help="the identification of the device",
 )
+shared_parser.add_argument(
+    "-k",
+    "--device-key",
+    type=str,
+    required=True,
```
thecode: Can we make this optional? I have checked and my devices doesn't require a key, we can add the logic in the API to use the key if user supplied it, if not just use "00".
The script can make this parameter optional with a default value, with this you can also just give a single example of usage with a key instead of modifying all examples

thecode:
In general the PR looks good, key can be added to all layers as implemented in this PR, however for the `control_device` script I suggest to make it optional with a default value of "00". This means that for users who don't need it, they don't need to learn what it is or change anything (if they are using the `control_device`) script.
HA will always fill this value regardless if it is needed or not, since this is not requiring any user intervention and very minor code changes.

I did only a quick review of the code, code changes looks good

scripts/control_device.py
YogevBokobza: Sure. Yes, you are right we can make it optional for this script.
I will make the changes.

YogevBokobza:
> In general the PR looks good, key can be added to all layers as implemented in this PR, however for the `control_device` script I suggest to make it optional with a default value of "00". This means that for users who don't need it, they don't need to learn what it is or change anything (if they are using the `control_device`) script. HA will always fill this value regardless if it is needed or not, since this is not requiring any user intervention and very minor code changes.
> 
> I did only a quick review of the code, code changes looks good

Cool I will make the needed changes.
The Switcher dev suggested calling this variable login_session instead of device_key.
What do you think?

scripts/control_device.py
YogevBokobza: Changes requested done.

YogevBokobza:
All done.
Waiting for merge.

thecode:
> Cool I will make the needed changes. The Switcher dev suggested calling this variable login_session instead of device_key. What do you think?

Although I usually prefer to go with the manufacture suggested name, in this case I would leave it as you implemented. we already have `session_id` and in general sessions are used for something which is temporary and valid for a period of time will in this case the value is fixed for a device so "device key" or "device login key" is more appropriate.

Thanks for digging in and fixing this issue.

thecode (approved):
From my point this is good to go, tested on the devices I have but they don't require key.

YogevBokobza:
@TomerFi All good. Please proceed with the merge :)
I will handle the support code for HA and the API code.

TomerFi:
I'm sorry for not getting back to you sooner.
@YogevBokobza @thecode Help me out here, guys. I wanna merge this, is this a fix for the current version, `3.4.0` and should be based off and merged to `dev`? Or is it part of the _Support new token-based Runner firmware versions_ and should be based off and merged to `runner-support`?

YogevBokobza:
> I'm sorry for not getting back to you sooner. @YogevBokobza @thecode Help me out here, guys. I wanna merge this, is this a fix for the current version, `3.4.0` and should be based off merged to `dev`? Or is it part of the _Support new token-based Runner firmware versions_ and should be based off and merged to `runner-support`?

Yes it part of 3.4.0 fix continuing dev.
Not relating to runner and the other logic.

thecode:
As @YogevBokobza wrote, this is a bugfix for existing devices, needs to be merged to dev and I would directly release a new version after.

TomerFi:
> Yes it part of 3.4.0 fix continuing dev.
> Not relating to runner and the other logic.

Gottcha. Alright, I'm waiting for the CI to finish, and I'll merge this.
Do we need a new 3.4.1 release after right?

YogevBokobza:
> > Yes it part of 3.4.0 fix continuing dev.
> > Not relating to runner and the other logic.
> 
> Gottcha. Alright, I'm waiting for the CI to finish, and I'll merge this. Do we need a new 3.4.1 release after right?

Yeah

TomerFi (approved):

TomerFi:
Included in [3.4.1](https://github.com/TomerFi/aioswitcher/releases/tag/3.4.1).

Issue #705, opened 2023-08-12
Switcher breeze There are no buttons Assume on & Assume off


Roei639:
### What happened?

Hello, 
I installed a new breeze switcher and there are no Assume on & Assume off buttons in Home Assistant

### Module Version

Integration Switcher Home Assistant

### Device Type

Switcher Breeze

### Firmware Version

2.20

### Relevant log output

```shell
Version Home Assistant: 2023.8.2
Installation type: Docker
Pi4
Log is empty
```

TomerFi:
> no Assume on & Assume off buttons in Home Assistant

Sound like this should be open in https://github.com/home-assistant/core.

Roei639:
I opened in core.
Below is the answer:
According to the data we have from Switcher this device - SWTA7003 doesn't support assume state (called OnOffType)
Please create an issue at [aioswitcher](https://github.com/TomerFi/aioswitcher) maybe Switcher can supply an updated DB.

Link: https://github.com/home-assistant/core/issues/98369#issuecomment-1680365980

Roei639:
?

Roei639:
@TomerFi  ??

TomerFi:
I'm sorry, I don't have the capacity to handle this any time soon.

Roei639:
> I'm sorry, I don't have the capacity to handle this any time soon.

the problem is still there...

TomerFi:
I don't have this capacity, but I took another look. Based on what you wrote, I need to see how this can be resolved here.

> According to the data we have from Switcher this device - SWTA7003 doesn't support assume state (called OnOffType)
Please create an issue at [aioswitcher](https://github.com/TomerFi/aioswitcher) maybe Switcher can supply an updated DB.

What do you suggest we do if a device of type SWTA7003 doesn't support the assumed state option?

Roei639:
> I don't have this capacity, but I took another look. Based on what you wrote, I need to see how this can be resolved here.
> 
> > According to the data we have from Switcher this device - SWTA7003 doesn't support assume state (called OnOffType)
> > Please create an issue at [aioswitcher](https://github.com/TomerFi/aioswitcher) maybe Switcher can supply an updated DB.
> 
> What do you suggest we do if a device of type SWTA7003 doesn't support the assumed state option?

In the Switcher app it is possible

Roei639:
[
[...]

PR #749, opened 2024-01-17
fix: updated breeze codes and fix test

YogevBokobza:
## Description

## Checklist

## Additional information

> Anything else?

### DIFF
--- a/tests/test_api_tcp_client.py
+++ b/tests/test_api_tcp_client.py
@@ -299,7 +299,7 @@ async def test_breeze_build_swing_command():
     remote_manager = SwitcherBreezeRemoteManager()
     remote_7022 = remote_manager.get_remote("ELEC7022")
     command = remote_7022.build_swing_command(swing=ThermostatSwing.ON)
-    assert_that(command.command).is_equal_to("000000004e4543587c32367c33327c31352c31357c31352c34307c31357c54303043387c33307c30317c414241425b33305d7c423234443642393445303146")
+    assert_that(command.command).is_equal_to("000000004e4543587c32367c33327c31352c31357c31352c34307c31357c54303042457c33307c30317c414241425b33305d7c423234443642393445303146")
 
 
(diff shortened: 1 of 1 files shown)


YogevBokobza:
@thecode Do You have a way to validate the DB against the issue opened?

thecode:
While this is an updated database the relevant AC that was mentioned in #705 still not fixed
```json
    "SWTA7003": {
        "IRSetID": "SWTA7003",
        "OnOffType": 0,
        "IRWaveList": [
```

The `OnOffType` should be 1 for ACs that are using `Toggle` command (so that the device doesn't know if the AC is on or off and we add buttons to allow manual override of the device state.

I can't simply modify this file since the codes for the AC also needs to be changes.

While I have no objection to update the DB (and we should update it when we get a new one) I don't understand how it is not reflecting what the user see in Switcher App. If the App shows manual state override and user needs it (for ACs that has separate on of commands this is irrelevant anyhow so he wouldn't look for it) the file should also indicate the same.

thecode (approved):

YogevBokobza:
> While this is an updated database the relevant AC that was mentioned in #705 still not fixed
> 
> ```json
>     "SWTA7003": {
>         "IRSetID": "SWTA7003",
>         "OnOffType": 0,
>         "IRWaveList": [
> ```
> 
> The `OnOffType` should be 1 for ACs that are using `Toggle` command (so that the device doesn't know if the AC is on or off and we add buttons to allow manual override of the device state.
> 
> I can't simply modify this file since the codes for the AC also needs to be changes.
> 
> While I have no objection to update the DB (and we should update it when we get a new one) I don't understand how it is not reflecting what the user see in Switcher App. If the App shows manual state override and user needs it (for ACs that has separate on of commands this is irrelevant anyhow so he wouldn't look for it) the file should also indicate the same.

Good point. I will ask the Switcher dev and update back.

TomerFi:
[3.4.2](https://github.com/TomerFi/aioswitcher/releases/tag/3.4.2)

PR #751, opened 2024-01-19
chore: seperate api1 and api2 to devices

YogevBokobza:
## Description

> Testing the idea of having different classes unique per device instead of SwitcherType1Api and SwitcherType2Api

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@TomerFi  @thecode 

I am trying to figure out a simple way to implement the logic of determining device type.
Take a quick look at this PR code (it's partial just to see the idea)

The idea here is that instead of SwitcherType1Api and SwitcherType2Api we gonna have classes per device like SwitcherTouch, SwitcherRunner, SwitcherBreeze, etc.
Eventually, those new classes would call the existing one but add specific variables like device_type which stores the name of the type and for example is_need_token. Those classes could even expand the functions of the existing ones.

This implementation means that it won't break the existing SwitcherType1Api and SwitcherType2Api until we replace them with the new classes.
We can have a function that tells the right class to use per device type string (as seen at the PR)

What do you think?

thecode:
I don't understand the benefits of this change. You now end up with devices that use the same API but needs to be handled differently. Currently for example we can handle all Thermostat devices (v2, touch, v4) using the same code, after this change which duplicates code here you will also need to duplicate the code in HA so each device will have its own class while they have an identical API.

I assume you want this to be for preparation for the token, but I don't see how it helps. Switch Touch & V4 (at least these two which I know) can be both with and without token.

I think it would be better if we first finalize the discussion at https://github.com/TomerFi/aioswitcher/discussions/715 before we try to make architectural changes

YogevBokobza:
> I don't understand the benefits of this change. You now end up with devices that use the same API but needs to be handled differently. Currently for example we can handle all Thermostat devices (v2, touch, v4) using the same code, after this change which duplicates code here you will also need to duplicate the code in HA so each device will have its own class while they have an identical API.
> 
> I assume you want this to be for preparation for the token, but I don't see how it helps. Switch Touch & V4 (at least these two which I know) can be both with and without token.
> 
> I think it would be better if we first finalize the discussion at #715 before we try to make architectural changes

I will try to explain in the discussion.

PR #761, opened 2024-04-04
chore: unite convert_str_to_devicetype inside devicetype class and move tests

YogevBokobza:
## Description

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@thecode @TomerFi 
I thought of a cleaner way to handle the DeviceType string to the DeviceType class.
Now it all sits under the "DeviceType" class which makes more sense than a "tool" function.

I will make the same kind of logic for the DeviceToken class which will include 2 more functions: one is to make sure the token is needed and valid (returns boolean true/false) and the other one is to call the tok.py script to convert the token into a packet (returns string with packet/empty string/rise error).

What do you think?

thecode:
I don't understand the reason behind this PR, passing device type as str is only used by the control device script, why do we want it as part of the device class? when calling the API we should use the Enum and not string

YogevBokobza:
> I don't understand the reason behind this PR, passing device type as str is only used by the control device script, why do we want it as part of the device class? when calling the API we should use the Enum and not string

You are right it's only for the control device script, but I thought this is a better place to put it.
I thought the tools function was not the right place for that..

Plus if a new device is added (Like the Runner S11) so it's one common place/file to add it and one common test file to add it.

Don't you think this is a better place to have it? like DeviceType handles all the stuff regarding the device type.

TomerFi:
I'm sorry I don't see any point in converting the `convert_str_to_devicetype` function to a static method.

YogevBokobza:
@thecode @TomerFi 
I can see you are both against this PR.. I am closing it.

PR #764, opened 2024-04-05
chore: updated validate token api and small refactoring

YogevBokobza:
## Description

> Took small changes from #747 and commit them as a new PR for this to be easier to review.

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@TomerFi
Can you exclude that secret from GitGuardian Security Checks?
It's just a string stored in Switcher just another simple gate.

thecode (dismissed):

YogevBokobza:
> @TomerFi Can you exclude that secret from GitGuardian Security Checks? It's just a string stored in Switcher just another simple gate.

@TomerFi ?

README.md
```diff
@@ -15,6 +15,12 @@ pip install aioswitcher
   <td><a href="https://github.com/TomerFi/aioswitcher/blob/dev/CONTRIBUTING.md">Contributing</a></td>
 </table>
 
+<strong>
+NOTE: For newer Switcher devices such as: Runner S11 you need to have a Token for communicating with devices.
+</strong>
+
+You can get it here: https://switcher.co.il/GetKey or by using scripts/get_token.py script
+
```
TomerFi: This doesn't fit here. Please think of a better place for it. I suggest using the docs site.

(previous cr https://github.com/TomerFi/aioswitcher/pull/747/files#r1551868132)

scripts/get_device_login_key.py
```diff
@@ -26,27 +26,30 @@
 _examples = """example usage:
 
 python get_device_login_key.py -i "111.222.11.22" -p 10002\n
+python get_device_login_key.py -i "111.222.11.22" -p 20002\n
+python get_device_login_key.py -i "111.222.11.22" -p 10003\n
+python get_device_login_key.py -i "111.222.11.22" -p 20003\n
```
TomerFi: One example should suffice. You can add a comment like "change the port based on your device" or something similar. But we don't need an example per port.

scripts/get_device_login_key.py
```diff
@@ -26,27 +26,30 @@
...
 """  # noqa E501
 
 parser = ArgumentParser(
-    description="Discover and print info of Switcher devices",
+    description="Get your login key of your Switcher devices",
```
TomerFi: "Get *the login key of your Switcher *device".
your = the
devices = device

src/aioswitcher/device/tools.py
```diff
@@ -26,6 +26,8 @@
 
 logger = getLogger(__name__)
 
+TOKEN_SECRET = "646a4d7335ea63cf9795c22ace1d6f957916b8be"
+
```
TomerFi: What's this?

src/aioswitcher/device/tools.py
```diff
@@ -165,11 +167,11 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
 
-def get_token(username: str, password: str) -> DeviceToken:
-    """Make API call to get a Token by username and password."""
+def get_token(username: str) -> DeviceToken:
+    """Make API call to get a Token by username."""
     token = DeviceToken("")
-    request_url = "https://switcher.co.il/GetKey/GetTok/t.php"
-    request_data = {"username": username, "password": password}
+    request_url = "https://switcher.co.il/GetToken/"
+    request_data = {"email": username, "secret": TOKEN_SECRET}
```
TomerFi: I don't understand this. Email is the only requirement for fetching the token from Switcher?!

TomerFi (changes requested):
Please also add the scripts to the README.

(previous cr: https://github.com/TomerFi/aioswitcher/pull/747#pullrequestreview-1980240859)

TomerFi:
Following https://github.com/TomerFi/aioswitcher/discussions/715#discussioncomment-9031235, please rebase.

README.md
YogevBokobza: I removed that and placed that in the Wiki..

src/aioswitcher/device/tools.py
YogevBokobza: A secret Switcher is set from their side to be alongside the Email of the user in the API call to get the token.
It's just so it will be sure that the call was from this library.

src/aioswitcher/device/tools.py
YogevBokobza: @TomerFi Yes.. Email is enough for them. This token part is only local internal communication with the devices. You can't do anything with this token if you are outside the user's network.

scripts/get_device_login_key.py
YogevBokobza: I did what you asked.

scripts/get_device_login_key.py
YogevBokobza: Done.

scripts/get_device_login_key.py
```diff
@@ -26,27 +26,30 @@
...
     formatter_class=RawDescriptionHelpFormatter,
 )
+parser.add_argument(
+    "-i",
+    "--ip-address",
+    required=True,
+    help="the ip address assigned to the device",
+    type=str,
+)
+possible_ports = [10002, 20002, 10003, 20003]
```
thecode: @YogevBokobza please try to avoid changes that are not requested during the review 
https://github.com/TomerFi/aioswitcher/discussions/715#discussioncomment-9018883 I had no comments from my side (PR was approved) and now we go step back. This can be easily added in a new PR after this one was approved. 

Please use port constants from https://github.com/TomerFi/aioswitcher/blob/acc2d035383f8998feffd5409fe9e0adfd6f75d0/src/aioswitcher/bridge.py#L48-L53

YogevBokobza: @thecode I am trying..
Anyway, what you asked for is done.

thecode (dismissed):
LGTM once comments from @TomerFi are resolved

YogevBokobza:
@TomerFi ping.. your attention is required.

TomerFi:
> @TomerFi ping.. your attention is required.

Also, being more patient is required.
I'm getting notifications for everything happening here, and I'll reply once I have the time; there's no need to re-ping whenever you think I'm not working fast enough for you. As we repeatedly told you, this is volunteer work; you may sometimes have to wait "a little" more than a few hours, especially during weekends.

src/aioswitcher/device/tools.py
TomerFi: But putting this here means it's no longer a secret, and anyone can use it from whatever endpoint they want, right?

src/aioswitcher/device/tools.py
TomerFi: But I can take your email and get your token myself. What is the token good for if anyone who knows your email can get it?

scripts/get_device_login_key.py
```diff
@@ -21,32 +21,47 @@
...
+    SWITCHER_UDP_PORT_TYPE2,
+    SWITCHER_UDP_PORT_TYPE2_NEW_VERSION,
+)
+
 printer = PrettyPrinter(indent=4)
 
 _examples = """example usage:
 
 python get_device_login_key.py -i "111.222.11.22" -p 10002\n
+
```
TomerFi: I'm not trying to be a stickler for details, but I think this redundant line break will affect the examples output by the argument parser.

src/aioswitcher/api/__init__.py
```diff
@@ -183,6 +183,7 @@ async def get_state(self) -> SwitcherStateResponse:
 
         Returns:
             An instance of ``SwitcherStateResponse``.
+
```
TomerFi, line 186: What's the point here? This adds line breaks in the code document.
The same goes for rows 246, 267, 360, and 710. I do not see any modifications in this file except for these five line breaks. Am I missing something? Was it because of some linting error?

TomerFi (changes requested):

src/aioswitcher/api/__init__.py
YogevBokobza: @TomerFi I just aligned those functions to the rest of the functions where there was an empty line right after the "Returns:" part

scripts/get_device_login_key.py
YogevBokobza: I don't know what you mean. this is working as should and also already working in the other scripts like: https://github.com/TomerFi/aioswitcher/blob/dev/scripts/control_device.py#L39-L62

src/aioswitcher/device/tools.py
TomerFi: If Switcher decides to take this insecure approach, that's on them. From our point of view, it would make more sense to get this token from users.
We can tell users, in the documentation, that they can use this value specifically when they invoke the function, but we should not store it in our code and inject it into the request without users knowing about it. It's important to relay this to users.

TomerFi: @thecode I'm interested in your take on this and the feasibility of passing the token as an argument from HA's point of view.

src/aioswitcher/device/tools.py
TomerFi: We can resolve this once we come to terms here: https://github.com/TomerFi/aioswitcher/pull/764#discussion_r1555854445.

src/aioswitcher/device/tools.py
YogevBokobza: Even if you have the token by the user's email you still need access to his network to be able to use it.
It is just another layer of security in the way.

Right now this [GetKey](https://switcher.co.il/GetKey) API URL exists and the token is sent to the user's email.
I believe @thecode requested a nicer way to get the token in HA by code semi-automatically and not a manual step outside HA and that is what this internal API is for.

So the scenario for HA is one of those:
1. Wait for the user to visit [GetKey](https://switcher.co.il/GetKey) API, enter his email, wait for the token to arrive in his email, input that into the HA prompt
2. A prompt in HA will ask the user to enter his email, the calling to get the token will be done from this library and HA can continue and use it.

thecode: I disagree with that, if you don't want the process of getting a token to be automatic, don't add support to the library at all. I don't care if Switcher decided this way as I didn't think the token protect anything anyhow. If someone is in your internal network the token doesn't worth anything.
I don't understand why Switcher changed their approach from using email + password to a fixed hardcoded based token but that is not our problem.

YogevBokobza: I will ask Switcher if it's possible to bring back the password alongside the email in the API call..

scripts/get_device_login_key.py
TomerFi: No, it's not.

Your change: 
image

The source you're basing on:
image

src/aioswitcher/api/__init__.py
TomerFi: That's good, keeping consitenency - But I don't understand why this is related to this PR.
This file wasn't event modified in this PR at all, besides these 4 line breaks in the code docs.

src/aioswitcher/device/tools.py
YogevBokobza: I spoke with Switcher, and setting a password with the API call is not an option.
a token is supposed to be something local only and unique for a user, asking for a password means you are getting outside from local to remote which is worse security-wise.

So as I can see we have 2 options:
1. Using [GetKey](https://switcher.co.il/GetKey) API and make a pause prompt in HA until the user is getting the token from his Email, also removing the API stuff implemented in this PR.
2. Using this API call as it is in this PR.

If you have a better way you can offer.. That is the Switcher answer..

scripts/get_device_login_key.py
YogevBokobza: Now I see what you mean.. fixed.

src/aioswitcher/device/tools.py
TomerFi: For the record, I have been voting for option one since we first discussed this "get token" script and module feature. Token retrieval is personal and a one-time requirement. Adding a "wrapper API feature" seemed redundant and too intrusive, in my opinion.

As for option two, I suggested passing the token as an argument, not dropping the function altogether. But I still prefer option one.

src/aioswitcher/api/tok.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def he(token: str | None) -> str:
+    if token is None:
+        raise RuntimeError("received None token as input")
+    encrypted_value = b64decode(bytes(token, "utf-8"))
+    cipher = AES.new(secret_key, AES.MODE_ECB)
+    decrypted_value = cipher.decrypt(encrypted_value)
+    unpadded_decrypted_value = unpad(decrypted_value, AES.block_size)
+    return hexlify(unpadded_decrypted_value).decode()
```
TomerFi: Looks like you added this file in https://github.com/TomerFi/aioswitcher/pull/764/commits/f5a37389fafd578cc56fd40d750962d1c2f7d575 addressing change requests.

Is this required for this PR?

YogevBokobza: Oh crap that was a mistake.. This will be in the next PR.. I will remove that now

YogevBokobza: File removed.

src/aioswitcher/device/tools.py
thecode: > I spoke with Switcher, and setting a password with the API call is not an option.

This was already implemented, what would have happened if we released a version and now they change drop the API? I don't understand this. This PR is tagged as "chore:" but it looks like it changes functionality.

Why did Switcher suddenly changed this? How exactly are you going to implement this in HA? I am starting to get tired of all these changes. You can pass my answer to Switcher that this is not acceptable, they constantly change things, add ports, change firmware requirements without considering that they have other users for their hardware/api.

@TomerFi where do you expect the users to get the token argument as you suggested? It won't be accepted to be added to HA docs.

@YogevBokobza the implementation in HA is going to be complicated anyhow, your current HA PR there will not be accepted, you will have to implement reauth flow and with this manual step it is going to be much harder.

I also remind you that the reason we asked to automate this step was that we don't have a way to validate the token which was my main concern in https://github.com/TomerFi/aioswitcher/discussions/715 and was mitigated by making this process automated. If we go back to a manual step lets go back to the discussion and reagree on how we implement the token. Sorry if it might disappoint you, but if Switcher is not sure how their solution should work and doesn't provide a reliable way to indicate if a token is valid I feel like it should not be supported.

YogevBokobza: > > I spoke with Switcher, and setting a password with the API call is not an option.
> 
> This was already implemented, what would have happened if we released a version and now they change drop the API? I don't understand this. This PR is tagged as "chore:" but it looks like it changes functionality.
> 
> Why did Switcher suddenly changed this? How exactly are you going to implement this in HA? I am starting to get tired of all these changes. You can pass my answer to Switcher that this is not acceptable, they constantly change things, add ports, change firmware requirements without considering that they have other users for their hardware/api.
> 
> @TomerFi where do you expect the users to get the token argument as you suggested? It won't be accepted to be added to HA docs.
> 
> @YogevBokobza the implementation in HA is going to be complicated anyhow, your current HA PR there will not be accepted, you will have to implement reauth flow and with this manual step it is going to be much harder.
> 
> I also remind you that the reason we asked to automate this step was that we don't have a way to validate the token which was my main concern in #715 and was mitigated by making this process automated. If we go back to a manual step lets go back to the discussion and reagree on how we implement the token. Sorry if it might disappoint you, but if Switcher is not sure how their solution should work and doesn't provide a reliable way to indicate if a token is valid I feel like it should not be supported.

The code before was communicating their test environment and not production environment and it was for progress and tests purposes.
working with password is not allowed in production real users environment.

I think you both are kind of forget something.. WE are integrating to them and not the other way around.
As far as Switcher, everything working just fine in their native app.
We need to thank them for their communication and they are trying to adapt things that fit us.
There are companies that blocks any other possible integration with HA by all mean.

I don't get why this current state going to be a problem in HA?
We can identify if the user needs a token and we can just prompt to enter email and from there it will be automatically.
What I am missing ?

thecode: > The code before was commenting their test environment and not production environment and it was for progress and tests purposes. working with password is not allowed in production real users environment.

Really? that was never mentioned, so they provided a solution they knew wouldn't work? I have no words.

> I think you both are kind of forget something.. WE are integrating to them and not the other way around. As far as Switcher, everything working just fine in their native app. We need to thank them for their communication and they are trying to adapt things that fit us. There are companies that blocks any other possible integration with HA by all mean.

If this is the approach I don't mind dropping myself from review and HA codewoner. They called me in the past and ask if I can help with integrating their devices into HA, I don't care if it works on their app, they should be thanking people working free for supporting their devices in other platform. Switcher is not the only manufacture I support in HA, and I can say that at most vendors alert me before on API changes (and apologize for having to change the API), provide good documentation for how their devices should be communicated and also take feedback and change the firmware according to our needs.

> I don't get why this current state going to be a problem in HA? We can identify if the user needs a token and we can just prompt to enter email and from there it will be automatically. What I am missing ?

If it is going to be an automated process as you described here it is OK, if you need the user to pass the API token or copy token from the email it is another solution, reading the discussion above it was mentioned about removing the automated token retrieval.

YogevBokobza: I was aiming in this PR to be fully automated get token in HA by api call function written in this library.
In HA the user only has to say it's email address.
@TomerFi think that this should be manually outside HA context, I don't think like him.

it's your both call now. I said and commited here the preferred way in my thinking.

TomerFi: > @TomerFi where do you expect the users to get the token argument as you suggested? It won't be accepted to be added to HA docs.

@thecode, I don't think this should be accepted into our docs either. This is not even a token per se; it's a "global password." And we are expected to store this "global password" in our code? What happens if it changes? Are we going to make releases for this "token"?

@YogevBokobza, I'm sorry, but this is not an acceptable approach. This "token" cannot be stored in the code!
I wasn't even on board with this approach when it was a password-based API function, but you both were OK with it, so I went along—but this: using a "global password"?! And storing it as a literal in the code?! This is not a solution for this project. This is an internal patch to run in a private environment.

---

@YogevBokobza, I have to ask: Why are you fighting us over this anyway? Why is it so important to automate the "token retrieval" process? It's a one-time action, no? What's wrong with users manually fetching their tokens in the old-fashioned way? This is how the vendor, Switcher, designed the process, no?
Instead of working on actually adding support for the new devices, we spent all of our time fighting you over stuff that makes sense only to you. Am I missing something?

@thecode, Am I missing something from HA's perspective? Is manually feeding tokens/passwords for users configuring integrations in HA unacceptable? Does HA expect this private information to be discoverable for all integrations?

YogevBokobza: @TomerFi I am not trying to fight you.
I mentioned more than once there are  2 ways go with that.
You are preferred the first way with manual fetching and @thecode is preferred the second one.
I am trying to make a progress here as @thecode said that the first way is not acceptable in HA and we need to get back to the board for that.
I am trying to continue but it's very hard that you too have different thoughts and block each other.
I am really don't know what to do and what to say anymore regarding this case..

YogevBokobza: Plus personally always prefer automatic way to get things and less manual steps as possible. That's how I am in my work as well..

Now that I am thinking about that, you said what if the secret would change? You can say that on manual step where you are waiting for an email that never comes for a reason like a bug.. at least in the automatic way we can detect the bug right away and handle that

Also what do you mean by "token cannot be stored in the code"?
How do we save device id in the code? device ip?
Again this token is local only and unique for the user and never change.

YogevBokobza: Also what do you mean by "token cannot be stored in the code"?
How do we save device id in the code? device ip?
Again this token is local only and unique for the user.

TomerFi: > I mentioned more than once there are 2 ways go with that.
You are preferred the first way with manual fetching and @thecode is preferred the second one.
I am trying to make a progress here as @thecode said that the first way is not acceptable in HA and we need to get back to the board for that.

We have a severe gap in communication here.
I don't like either approach. Tokens and passwords are personal. The vendor created a way for users to get their tokens online. You are jumping through hoops to automate something that wasn't meant to be automated to begin with. And I would be surprised if @thecode "likes" an approach that saves a secret value as literal in the code. I'm guessing you misunderstood.

> I am trying to continue but it's very hard that you too have different thoughts and block each other.
I am really don't know what to do and what to say anymore regarding this case..

This is called a review, if the reviewer doesn't like/understands something you are trying to contribute, if a reviewer doesn't see eye to eye with you on something, the contribution is blocked until you and the reviewer can come to terms eachother. This is just how reviewing works.

> Plus personally always prefer automatic way to get things and less manual steps as possible. That's how I am in my work as well..

Me, too. But this is not automation. This is putting comfort over security and privacy. If I'm downloading an app and it asks for a password for one of my devices, I would do some research on the app before handing out my private information. If the app says, "No, just give me your email, and I can figure out your password myself," there's a good chance I will first disconnect my network.

I'm still hoping to answer my question: Why is this "automation" so important to you? Why are you blocking yourself over it? What's wrong with continuing to work on integrating the device and letting users fetch their tokens from Switcher?

> Now that I am thinking about that, you said what if the secret would change? You can say that on manual step where you are waiting for an email that never comes for a reason like a bug.. at least in the automatic way we can detect the bug right away and handle that

- How exactly do you expect to handle it? What can you do if the vendor doesn't return the token?
- Why would you want to handle it? This is between users and the vendor, why do you want to get involved? What happens if tomorrow Switcher opts for a more secure approach (which they will probably do eventually) and your code won't work anymore?

I honestly can't see any reason for us to do it unless it's mandatory for HA, which I don't think it is (@thecode, correct me if I'm wrong).

> Also what do you mean by "token cannot be stored in the code"?
How do we save device id in the code? device ip?

Where exactly are we storing device IPs or personal IDs in our code?

> Again this token is local only and unique for the user and never change.

No. This "token" is used to access the Switcher vendor API; the token returned from the vendor API request is used internally only.
And if it's unique and never going to change, then why are we even talking about this? Let users get their unique, never-changing token by themselves and feed it to the integration that would feed it to this module.

thecode: I think we can't progress anymore, sorry @YogevBokobza. For every step forward we take 10 steps back.

@TomerFi in short, there is no restriction in HA for doing the token retrieval manually, in fact many integrations ask you to type a code from the vendor app or a token you got in another way. 

However I haven't seen any integration util now that doesn't have a way to validate the token. This was my main concern in https://github.com/TomerFi/aioswitcher/discussions/715 a last resort mitigation was to automate the token retrieval process so that we reduce the change for incorrect token. While there is no restriction in HA, I will not approve a manual token solution, I don't mind stepping back and let another reviewer accept this in HA if this is the only way, but I don't want to face users typing (even if copy paste) a long token with no way to validate that it is correct. Just yesterday I tried to link an app that needed a token code and since they are on different devices I thought it would be easier to type it in stead of copy/paste which will need a shared method between them, with a token half length from what used by Switcher I had 2 mistakes (had to retype 3 times), however I got response that the token is invalid so I could check again.
The current implementation by Switcher is  that if the token is invalid the device just ignore the request. This create a situation that it will be impossible to detect if it is a network problem or a token problem.

YogevBokobza: @thecode I agree with your points why automatic way is preferred against manual step and yes you are right that we can't validate the token and we cannot detect if this is a network issue or token issue.

I have an idea but I am not sure if this would help us..
We could make this whole get token API as a private function and not accessible throught "tools", then in HA we ask the users when needed a token to type in their email address and to go and get their token manually using the URL that will send the token to their email box, then when they type in the token got from the email we can run the get token API and make sure that the result based of their email is the same as accessing and get their token by the email box.

How does it sounds?

TomerFi: Let's pause, step back, and discuss what's required because we're too scattered.

---

**Goal**

We want to support new Switcher devices that require tokens. The tokens are unique and immutable per email address and generated by the Switcher.

---

**Problem**

Switcher devices that require tokens—correct me if I'm wrong—please Do not respond to requests with the wrong or no token in a manner that will allow us to differentiate network errors from token errors.

Home Assitant won't accept integration with devices that do not differentiate network and token errors. The main concern is that users might fail while pasting/typing their tokens and won't be able to identify the problem type.

---

**Questions**

1. Do we know if the Switcher intends to change this situation? Are there any plans to return an "unauthorized" response for failed tokens?

---

**Proposed Solutions**

**Proposed Solution 1 - Why even bother**

Honestly, I don't know what the big deal is. Given that the token is unique and immutable. This is a "one-time action" done at the start of the user's journey to configure Switcher integration in Home Assistant. It's different from devices that suddenly won't accept the tokens. Token errors can only occur during the integration setup if I understand correctly. If we put a statement in bold in the integration documentation, "Please make sure the token you set in Home Assistant is identical to the token Switcher sent you" (or something of that sort), it should suffice, in my opinion, if not, the word spread after the first 3-4 forum discussions.

**Proposed Solution 2 - Validate token with a designated API**

Let's ask Switcher to expose a designated API endpoint that takes an email and a token and returns whether or not the given token is valid for the provided email address. Home assistant users will feed their emails and tokens to the integration configuration, and we will use the designated API endpoint to validate the given info.
Whether the validation should be handled at the integration or module levels is debatable.

This is the most straightforward, efficient, and correct approach to validating the token in the manner we require for this integration.

**Proposed Solution 3 - Validate the token with a regex expression**

Let's ask Switcher for a regex expression matching their token generation process and validate it locally. If they prefer, they can give us code for a utility function to validate the token instead of a regex expression, but this cannot be unreadable, questionable, "black box" code. A code contribution to this repository must adhere to the project's standards, and it doesn't matter who created it.

This solution can be combined with the previous one. We can validate locally before sending a validation request to Switcher, reducing network usage.

**Proposed Solution 4 - Fetch the token on behalf of the users**

The current solution.

**Proposed Solution 4.1 - Users will feed email and password**

Switcher exposes an API endpoint taking email and password and returning the token. This is the previous iteration of the `get_token` functionality. Users feed email and password to the integration, we fetch the token on their behalf, and use it.

The previous suggestions are much better, in my opinion, but I can live with this one if that's the only thing we can agree on (it has already been accepted and merged).
However, for some reason, I don't understand. Home Assistant won't accept this approach. Not that it matters because Switcher won't support this approach in production. The latter actually makes sense to me. Switcher is probably trying to avoid the overhead of handling passwords.

**Proposed Solution 4.2 - Users will feed email**

As an alternative to 4.1, based on this PR, Switcher gave us one password meant to be used by this module instead of handling user passwords. With this password, we can get any token for any email, such as a "free pass," if you will.

It's not the ideal solution, but I could have lived with that if there had been any way to keep the password secret. Storing the password as a literal inside the code is unacceptable!
What are we even trying to achieve here? Switcher make the right call and decides to make their devices more secure, so are we trying to make them less secure?

To bridge the gap, I suggested passing the password as an argument. But Home Assistant won't accept this, and it makes sense. They don't want to store the password in the docs. I can relate; I don't want to store the password in the code.

I'm gonna take a wild guess here, I assume this is the only solution acceptabile in Home Assistant, because we hide the bad part in this module. You can push something like that directly to Home Assistant, such as some code where you store the password as a literal, and see where it gets you.

---

We have enough alternatives here to keep going forward. I can understand if you don't agree. In that case, the best thing for this project will be to postpone this work until Switcher changes something we can agree on.

src/aioswitcher/device/tools.py
thecode: @TomerFi just a comment
>Proposed Solution 1 - Why even bother

The details there are incorrect, Many users already have the integration active, we can't break it for them if they have no action to do. HA allow asking users to provide new auth details (called reauth flow) when integration detects that it needs new data from the user. This is mainly used if you change a password to login to your device or cloud based API.
My plan was that when we detect a device that require a token we ask users to add the token, if we don't detect such device we never ask anything from the user. The problem is that only the integration can trigger this, the user can't trigger this flow. So what will happen is that if we trigger this, a user supply a token, we don't know how to validate it, we just store it and later the user finds that he can't control his devices. This will immediately goes into an issue "Why is my device not working" it doesn't matter if you document this in font 36, there is no option for the user to understand where is the problem.

.. but lets assume the user realized he typed a wrong token, the only way for him to fix it now is to remove the integration and add it again, this also removes any entity customizations he did (names, icons) and delete all related automations. This would bring a very bad user experience 

3 in my opinion wouldn't be possible or bring anything.

2 & 4 are acceptable by me (with 4.1 as the most preferred solution, since it can only happen with the user who has the details for his account and if you have them it is the same as you do it from their APP (if I have you user and password I can install the APP on my phone and do it.) the benefit here is that you don't store these details, you ask them from the user get the token and only store the token. The token can not be used later to get the user APP details.

4.2 is useless, I really don't understand Switcher, they have now officially provided anyone a backdoor for their Token logic, anyone who has the request token can ask a token for any valid email, so even if they drop this API, I assume there are already many breached tokens. 

>I assume this is the only solution acceptable in Home Assistant, because we hide the bad part in this module

At first I thought we can put the token here, but when you bump the lib the changes are reviewed, I would not want to be the one who approve such change since we all realize it is insecure. 

but considering the history with Switcher I must say I have low hopes, they provided an API (4.1) and changed it while we where working on it, I am also spending too much time for this, I don't blame anyone, but I would not be able to keep this pace for long. I was aiming to give a push this 2 weeks so we finish this, but it doesn't look like it is goin to happen.

TomerFi: @thecode 
> The details there are incorrect...

I appreciate the clarification. I suggest that Home Assitant needs some modifications, i.e., not removing the entire integration for modifying tokens. But I assume there's a reason for that, and changing it is more challenging than it sounds. Regardless, I understand and see your point.

> 3 in my opinion wouldn't be possible or bring anything.

It can help verify some typos, at the very least. But I agree with you; after you clarify the requirement, this solution needs more merit.

> 2 & 4 are acceptable by me (with 4.1 as the most preferred solution

Yep, I agree with you. Option 2 should be the preferred option; if it is not possible, option 4.1 is the next. Other options should be off the table.

> 4.2 is useless, I really don't understand Switcher. They have now officially provided anyone a backdoor for their Token logic.

Yeah, I was shocked at this, too. It's like we care about the security and privacy of their devices more than they do. Part of me wants to believe that maybe they didn't think it'd end up as literal in our code base.

---

@YogevBokobza, our next step is to check with Switcher for their preference. Either they provide us with an endpoint for verifying email and tokens (option 2) or an endpoint for fetching tokens per user and password (option 4.1). We're good with both, but prefer the former.

src/aioswitcher/device/tools.py
TomerFi: We accidentally continued this discussion in the next change request; please do not reply to this one anymore to avoid confusion.
Next comment: https://github.com/TomerFi/aioswitcher/pull/764#discussion_r1560181867

src/aioswitcher/device/tools.py
YogevBokobza: > Do we know if the Switcher intends to change this situation? Are there any plans to return an "unauthorized" response for failed tokens?

I believe that Switcher does not intend to change this situation. They return "Invalid credentials" with the current implementation API as you can see added to the tests.

Option 2 I feel like it's much like the option I suggested here in the [comment](https://github.com/TomerFi/aioswitcher/pull/764#discussion_r1559267304)
I think it's doable, I will reach Switcher for that.

Option 4.1 is too risky for them, I don't see it happening.

If option 2 is done by Switcher we can continue? Can we reach this new API in this library?

YogevBokobza: So based on option 2 Switcher made that new API call.
I made the relevant changes converted get_token to validate_token with docs and tests and removed the secret.
@TomerFi @thecode Please see if that's ok and we can continue :)

TomerFi (approved):
LGTM
Thank you for your patience. I think we can merge this; I would just prefer if @thecode could take a look before as well.

As a side note, I don't see why we need the validate-token script, but it can be helpful for troubleshooting.

src/aioswitcher/device/tools.py
TomerFi: I'm waiting for @thecode to have a look before resolving this.

YogevBokobza:
> LGTM Thank you for your patience. I think we can merge this; I would just prefer if @thecode could take a look before as well.
> 
> As a side note, I don't see why we need the validate-token script, but it can be helpful for troubleshooting.

No problem, I am happy that we can have a solution that everyone is happy with.
Yeah that script is for troubleshooting and can be used for the WebApi repo as well.

thecode (approved):

src/aioswitcher/device/tools.py
YogevBokobza: @TomerFi Hallelujah 😃 
You can resolve and merge :)

TomerFi:
@thecode @YogevBokobza Thank you!

PR #766, opened 2024-04-12
chore:  added token support in types and a convertion utility function

YogevBokobza:
## Description

> Adding the tok.py had in the previous s11 PR, added token needed for the devices in the discovery and added tests.

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@thecode @TomerFi 
Do you have any idea why the tests are failing for Python 3.9? 
For Python 3.10 it's working fine.

YogevBokobza:
> @thecode @TomerFi Do you have any idea why the tests are failing for Python 3.9? For Python 3.10 it's working fine.

Never mind. I think I fixed that [here](https://github.com/TomerFi/aioswitcher/pull/766/commits/baea774c51ab2643e9b223609f142f84b454a081) and [here](https://github.com/TomerFi/aioswitcher/pull/766/commits/ddd2cbe7d660d46d27630f6bbf77edb0f96b834c)

src/aioswitcher/device/tok.py
```diff
@@ -0,0 +1,35 @@
...
+"""Switcher convert token to valid packet string. Input: Token (String), Output: Packet value (String)"""
+
+from base64 import b64decode
+from binascii import hexlify
+from typing import Union
+
+from Crypto.Cipher import AES
+from Crypto.Util.Padding import unpad
+
+secret_key = b"jzNrAOjc%lpg3pVr5cF!5Le06ZgOdWuJ"
```
TomerFi: Let's not call it "secret" since it's not. Something like `token_key` would be better. However, we only use it in the function below, so do we really need it as a variable?

src/aioswitcher/device/tok.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def he(token: Union[str, None] = None) -> str:
```
TomerFi: Please change the name to something more meaningful and add code docs.

src/aioswitcher/device/tok.py
```diff
@@ -0,0 +1,35 @@
+# Copyright Tomer Figenblat.
```
TomerFi: Why do we need a file for one function? Can't we put this function in the `tools.py` file with the other utility functions?

pyproject.toml
```diff
@@ -41,6 +41,7 @@ pytest-mockservers = "^0.6.0"
 pytest-sugar = "^0.9.4"
 time-machine = "^2.7.0"
 yamllint = "^1.26.3"
+pycryptodome = ">=3.18.0"
```
TomerFi: This is the development dependencies group. We need to bump this to `[tool.poetry.dependencies]` if we want `pycryptodome` to be available for the token function outside the development environment.

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +166,34 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+    if is_token_needed:
+        if is_token_not_empty:
+            try:
+                token_packet = tok.he(token)
+            except (KeyError, ValueError) as ve:
+                raise RuntimeError("convert token to packet was not successful") from ve
+            return str(token_packet)
+        else:
+            raise RuntimeError("a token is needed but missing")
+    return ""
```
TomerFi: Why are we returning an empty string here? Can't we return `None` instead?

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +166,34 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
     return DeviceType.MINI
 
 
+def convert_token_to_packet(
+    device_type: DeviceType, token: Union[str, None] = None
+) -> str:
+    """Return token packet from token if is valid, otherwise empty string."""
+    is_token_needed = bool(device_type.token_needed)
```
TomerFi: Isn't `device_type.token_needed` already a boolean? why are you casting it?

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +166,34 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+    is_token_not_empty = token is not None and str(token) != ""
```
TomerFi: Instead of this: `token is not None and str(token) != ""`, try this: `bool(token)`.

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +166,34 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+
+
+def is_token_valid(device_type: DeviceType, token: Union[str, None] = None) -> bool:
+    """Return true if token is used and valid."""
+    is_token_needed = bool(device_type.token_needed)
+    is_token_not_empty = token is not None and str(token) != ""
+    if is_token_needed:
+        if is_token_not_empty:
+            return True
+    return False
```
TomerFi: Looks like this entire function can be replaced with a simple statement, why do we need it as a function?

```python
device_type.token_needed and bool(token)
```

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +166,34 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
     return DeviceType.MINI
 
 
+def convert_token_to_packet(
```
TomerFi: I'm just making sure we aren't missing anything. Currently, no one is using this function, only tests. This is meant to be used with future steps of this work, right?

src/aioswitcher/device/__init__.py
```diff
@@ -36,25 +36,31 @@ class DeviceCategory(Enum):
...
-    RUNNER_MINI = "Switcher Runner Mini", "0c02", 2, DeviceCategory.SHUTTER
+    MINI = "Switcher Mini", "030f", 1, DeviceCategory.WATER_HEATER, False
+    POWER_PLUG = "Switcher Power Plug", "01a8", 1, DeviceCategory.POWER_PLUG, False
+    TOUCH = "Switcher Touch", "030b", 1, DeviceCategory.WATER_HEATER, False
+    V2_ESP = "Switcher V2 (esp)", "01a7", 1, DeviceCategory.WATER_HEATER, False
+    V2_QCA = "Switcher V2 (qualcomm)", "01a1", 1, DeviceCategory.WATER_HEATER, False
+    V4 = "Switcher V4", "0317", 1, DeviceCategory.WATER_HEATER, False
+    BREEZE = "Switcher Breeze", "0e01", 2, DeviceCategory.THERMOSTAT, False
+    RUNNER = "Switcher Runner", "0c01", 2, DeviceCategory.SHUTTER, False
+    RUNNER_MINI = "Switcher Runner Mini", "0c02", 2, DeviceCategory.SHUTTER, False
```
TomerFi, line 47: I'm just making sure we aren't missing anything. All the current devices are set to `False.` I assume new devices contributed in the next steps of this work will be `True`, right?

TomerFi (changes requested):

src/aioswitcher/device/__init__.py
YogevBokobza: You are right. Switcher Runner S11 will be True.

src/aioswitcher/device/tok.py
thecode:
```suggestion
def convert_token_to_packet_data(token: Union[str, None] = None) -> str:
```
or something similar

src/aioswitcher/device/tok.py
thecode: I think it would fit in tools

src/aioswitcher/device/tools.py
YogevBokobza: Hmm.. Don't know the influence. I can try.

src/aioswitcher/device/tok.py
thecode: I would also consider just adding these lines inside `convert_token_to_packet` since it is used only there and only few lines

src/aioswitcher/device/tools.py
YogevBokobza: Good point I will see

src/aioswitcher/device/tools.py
YogevBokobza: Because I am calling this function many times when using the API calls.
I think it's better to write this once and just call it instead of adding this statement over and over in the code

src/aioswitcher/device/tools.py
YogevBokobza: Correct. In the next PR it will be used

src/aioswitcher/device/tok.py
YogevBokobza: Looks good I will refactor

TomerFi: This can be resolved together with https://github.com/TomerFi/aioswitcher/pull/766#discussion_r1563001097.

YogevBokobza: > I would also consider just adding these lines inside `convert_token_to_packet` since it is used only there and only few lines

Sorry but I didn't understand what you mean..

src/aioswitcher/device/tools.py
TomerFi: I'm afraid I have to disagree. It's new and not being used, so I don't see any point in having it as a function. What do you mean by `over and over`? How many junctions in this code do you expect will require this? This should be a lot to justify; it's just a simple ~40-charachters evaluation statement.

TomerFi: If you think you have enough usage for this function to justify it. Could you include it with the PR that will be used instead of here? I think there's no point in doing this early.

pyproject.toml
YogevBokobza: It is done.

src/aioswitcher/device/tok.py
YogevBokobza: It is done. I moved it under `tools.py` and removed the file.

src/aioswitcher/device/tok.py
YogevBokobza: It is done I put the tok.py logic under `convert_token_to_packet`

src/aioswitcher/device/tok.py
YogevBokobza: It is done.

src/aioswitcher/device/tok.py
YogevBokobza: It is done.

src/aioswitcher/device/tools.py
YogevBokobza: It is done. I will see if it will make an issue in the next PR with the usage of that.

src/aioswitcher/device/tools.py
YogevBokobza: It is done.

src/aioswitcher/device/tools.py
YogevBokobza: It is done.

src/aioswitcher/device/tools.py
YogevBokobza: I think there is enough usage in the next PR with the full RUNNER S11 integration and the next devices in the plan.
Here are some examples of usage for the next PR:
https://github.com/TomerFi/aioswitcher/pull/747/files#diff-cec958238bf6f0279cfc946924d9d860586123564684aaa276671879bcee8195R173

https://github.com/TomerFi/aioswitcher/pull/747/files#diff-cec958238bf6f0279cfc946924d9d860586123564684aaa276671879bcee8195R192

https://github.com/TomerFi/aioswitcher/pull/747/files#diff-cec958238bf6f0279cfc946924d9d860586123564684aaa276671879bcee8195R777
 
I think it makes the code that uses it cleaner..

src/aioswitcher/device/tools.py
```diff
@@ -168,30 +171,40 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+        Token packet if token is valid,
+        otherwise empty string or raise error.
+
+    """
+    token_key = b"jzNrAOjc%lpg3pVr5cF!5Le06ZgOdWuJ"
+    is_token_needed = device_type.token_needed
+    is_token_not_empty = bool(token)
     if is_token_needed:
-        if is_token_not_empty:
+        if is_token_not_empty and token is not None:
```
thecode: The logic here seems strange, you call a function `convert_token_to_packet` which should return a token and the function itself check if the token is needed, it makes no sense, the call should check if the token is needed and decide if it should call this function, you are going to build a different packet/flow if token is needed so there is no use to check this inside here since you need to do this anyway on the caller side.

src/aioswitcher/device/__init__.py
```diff
@@ -30,6 +30,7 @@ class DeviceCategory(Enum):
     POWER_PLUG = auto()
     THERMOSTAT = auto()
     SHUTTER = auto()
+    SHUTTER_SINGLE_LIGHT_DUAL = auto()
```
thecode: I think it will make more sense to call it 
```suggestion
    SINGLE_SHUTTER_DUAL_LIGHT = auto()
```

src/aioswitcher/device/tools.py
YogevBokobza: It's not that strange if you can see where I am using it. [here](https://github.com/YogevBokobza/aioswitcher/blob/a193f88e52369401adce86ac98bc00736ad31bdc/src/aioswitcher/api/__init__.py#L113)
In this place, I want to set the token to be empty or token packet and "stop" the user there if there is an error

thecode: It is not recommended to use a function that raise inside `__init__`, set the token there without any check, and when you create the packet raise if it is missing

thecode: and as I wrote earlier this check is redundant since you do it anyhow when you use the token:
https://github.com/YogevBokobza/aioswitcher/blob/a193f88e52369401adce86ac98bc00736ad31bdc/src/aioswitcher/api/__init__.py#L192

YogevBokobza: I see what you mean but the problem is that the convert function may raise an error when you are converting it if the token is not in the right format.. So I don't know at which time the conversion should happen.. 
Do you have an idea?

YogevBokobza: > and as I wrote earlier this check is redundant since you do it anyhow when you use the token: https://github.com/YogevBokobza/aioswitcher/blob/a193f88e52369401adce86ac98bc00736ad31bdc/src/aioswitcher/api/__init__.py#L192

They are not that much the same.. in the place you just showed we need to validate that the token is needed and not empty so we can use the next packet for the API and the first is to get the token packet that can be a string or None or raise an error so we can use that in the next packet for the API

How would you approach this?

YogevBokobza: Do you think that instead of [this](https://github.com/YogevBokobza/aioswitcher/blob/a193f88e52369401adce86ac98bc00736ad31bdc/src/aioswitcher/api/__init__.py#L113)
it will be better to not have `self._token` and have the token_packet as a normal variable that is defined in the operations API like `set_position` and use that instead of `is_token_valid`? (In this approach a conversation would have to happen every time..)

YogevBokobza:
@thecode @TomerFi 
I just found out [this](https://github.com/YogevBokobza/aioswitcher/blob/add-token-convert-with-discovery/src/aioswitcher/device/__init__.py#L228-L244) leftover from when we used that API call to get the token..
Do we want to have this token inside a `DeviceToken` class and possibly make the conversion there? or just have it as a string and remove this class?

src/aioswitcher/device/tools.py
TomerFi: I won't review a closed PR—it's closed for a reason. Please include this in the PR where you're using it if you think it's needed, and we can continue the discussion there.

For the record, I would like to clarify your request. You want to replace **3** occurrences of:
```python
device_type.token_needed and bool(token)
```

With **3** occurrences of:
```python
is_token_valid(self._device_type, self._token)
```
And this utility function:
```python
def is_token_valid(device_type: DeviceType, token: Union[str, None] = None) -> bool:
    "Return true if token is used and valid."""
    return device_type.token_needed and bool(token)
```

And this testing function:
```python
@mark.parametrize("type, token, is_token_valid", [
    (DeviceType.RUNNER, "", False),
    (DeviceType.RUNNER, "zvVvd7JxtN7CgvkD1Psujw==", False),
    (DeviceType.RUNNER, None, False),
    (DeviceType.RUNNER_S11, "zvVvd7JxtN7CgvkD1Psujw==", True),
    (DeviceType.RUNNER_S11, "", False),
    (DeviceType.RUNNER_S11, None, False)
    ])
def test_is_token_valid_should_return_expected_bool(type, token, is_token_valid):
    assert_that(tools.is_token_valid(type, token)).is_equal_to(is_token_valid)
```

---

I am trying to understand how this is considered cleaner.
The following is very readable to Python developers. If you think a clarification is in order, add a comment. 

```python
device_type.token_needed and bool(token) # this validates the token
```

YogevBokobza: I didn't want to add the usage of this because there are dependants on that and I didn't want the PR become bigger..
There are more than 3 usages I just mentioned those as an examples.

I think this is a better approach and also covert by tests so why not?

src/aioswitcher/device/tools.py
TomerFi: That's better. Thanks.

We now have:
```python
is_token_needed = device_type.token_needed
is_token_not_empty = bool(token)
if is_token_needed:
    if is_token_not_empty and token is not None:
```

Why do we need a variable named `is_token_needed` to store the value of `device_type.token_needed`, mainly if it is only evaluated once in the next two rows?

Also, Python's built-in _bool_ function already verifies the token `is not None`. (See [here](https://docs.python.org/3/library/functions.html#bool) and [here](https://www.w3schools.com/python/ref_func_bool.asp)).

You can change the above code to the following, and it will have the same effect:
```python
if device_type.token_needed:
    if bool(token):
```

Or, if you insist:
```python
is_token_not_empty = bool(token)
if device_type.token_needed:
    if is_token_not_empty:
```

---

But honestly, if you're trying to make this more readable, I would suggest reversing the `if` branches (please forgive any bugs, this was written in-comment):
```python
if not device_type.token_needed:
    return None

if not bool(token):
    raise RuntimeError("a token is needed but missing")

try:
    encrypted_value = b64decode(bytes(token, "utf-8"))
    cipher = AES.new(b"jzNrAOjc%lpg3pVr5cF!5Le06ZgOdWuJ", AES.MODE_ECB)
    decrypted_value = cipher.decrypt(encrypted_value)
    unpadded_decrypted_value = unpad(decrypted_value, AES.block_size)
    return hexlify(unpadded_decrypted_value).decode()
except (KeyError, ValueError) as ve:
    raise RuntimeError("convert token to packet was not successful") from ve
```

src/aioswitcher/device/tools.py
TomerFi: Discussion continues at https://github.com/TomerFi/aioswitcher/pull/766#discussion_r1567465804.

src/aioswitcher/device/tools.py
TomerFi: > I won't review a closed PR—it's closed for a reason. Please include this in the PR where you're using it if you think it's needed, and we can continue the discussion there.

src/aioswitcher/device/tools.py
YogevBokobza: > Why do we need a variable named is_token_needed to store the value of device_type.token_needed, mainly if it is only evaluated once in the next two rows?

I just thought that the naming made more sense and was readable but I can also not use the variables..

> Also, Python's built-in bool function already verifies the token is not None. (See [here](https://docs.python.org/3/library/functions.html#bool) and [here](https://www.w3schools.com/python/ref_func_bool.asp)).

I tried that.. but I was getting a fail by linting.. I had to add `is not None` directly so I could pass it.
That is also true for those:

> if device_type.token_needed:
>     if bool(token): 

and

> is_token_not_empty = bool(token)
    if device_type.token_needed:
       if is_token_not_empty:

As for that:

> But honestly, if you're trying to make this more readable, I would suggest reversing the if branches (please forgive any bugs, this was written in-comment):

I can try that..

src/aioswitcher/device/tools.py
YogevBokobza: > > I won't review a closed PR—it's closed for a reason. Please include this in the PR where you're using it if you think it's needed, and we can continue the discussion there.

I can't add the usage parts as I mentioned.. it will need to bring all Runner S11 stuff with that which makes this PR big.
I am trying to make this PR small as possible.. I am not sure how to progress..

src/aioswitcher/device/tools.py
TomerFi: What was the linting error?

src/aioswitcher/device/tools.py
TomerFi: I don't understand. Please clarify: Why can't you delete `is_token_valid` and its testing function from this PR despite them not being used at all?

TomerFi:
> @thecode @TomerFi I just found out [this](https://github.com/YogevBokobza/aioswitcher/blob/add-token-convert-with-discovery/src/aioswitcher/device/__init__.py#L228-L244) leftover from when we used that API call to get the token.. Do we want to have this token inside a `DeviceToken` class and possibly make the conversion there? or just have it as a string and remove this class?

I vote for removing the class. I don't see any value in encapsulating the token.

src/aioswitcher/device/tools.py
YogevBokobza: > But honestly, if you're trying to make this more readable, I would suggest reversing the if branches (please forgive any bugs, this was written in-comment):

I implemented the same right now, you will see in the status check the lining issue I am facing..

src/aioswitcher/device/tools.py
YogevBokobza: > I don't understand. Please clarify: Why can't you delete is_token_valid and its testing function from this PR despite them not being used at all?

I am not going to argue with you.. I removed the function and its test.

YogevBokobza:
> > @thecode @TomerFi I just found out [this](https://github.com/YogevBokobza/aioswitcher/blob/add-token-convert-with-discovery/src/aioswitcher/device/__init__.py#L228-L244) leftover from when we used that API call to get the token.. Do we want to have this token inside a `DeviceToken` class and possibly make the conversion there? or just have it as a string and remove this class?
> 
> I vote for removing the class. I don't see any value in encapsulating the token.

I removed that DeviceToken class.

src/aioswitcher/device/tools.py
YogevBokobza: > Do you think that instead of [this](https://github.com/YogevBokobza/aioswitcher/blob/a193f88e52369401adce86ac98bc00736ad31bdc/src/aioswitcher/api/__init__.py#L113) it will be better to not have `self._token` and have the token_packet as a normal variable that is defined in the operations API like `set_position` and use that instead of `is_token_valid`? (In this approach a conversation would have to happen every time..)

I think we should discuss this next PR when I will show you the usage of the convert token to token packet..

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +169,38 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+    Return:
+        Token packet if token is valid,
+        otherwise empty string or raise error.
+
+    """
+    if not device_type.token_needed:
+        return None
+
+    if not bool(token):
+        raise RuntimeError("a token is needed but missing")
```
thecode: This check should be done where you call this function, you don't need to call it, if you know the device doesn't need a token.
The rest of the function is very clear, it try to convert a token and raise if the conversion fails. but there is no reason to put into this function a device specific logic (the test if the device needs a token) which also require you to pass the device type here just for that, while I think you will need to do this check again where you use it since there are two different APIs anyhow (one with a token and one without) so when you call the API that needs a token you will do the test again.

thecode: In addition if you do the check where you intend to call the function after you realized you need a token you don't need to return `None` as you do here, you simply do
```python
    if not bool(token):
        raise RuntimeError("a token is needed but missing")
     
     convert_token_to_packet(...
```
 or if you want to to both checks at once (on the caller)
 
 ```python
    if device_type.token_needed and not bool(token):
        raise RuntimeError("a token is needed but missing")
     
     convert_token_to_packet(...
```

Comment (YogevBokobza) [reply src/aioswitcher/device/tools.py]: I got you but @TomerFi asked to remove the `is_valid_token` function so I thought this is the only place we can do this check.. plus don't you think we should avoid this conversation in every API call? That's why i intended to put this conversation with the check in the device init time

Comment (TomerFi) [reply src/aioswitcher/device/tools.py]: I agree with @thecode; it would be better to check if you even need a token before invoking the `convert_token_to_packet.`

But it has nothing to do with removing the `is_valid_token` function. Validity and requirement are two different things. If a token is not required, it doesn't necessarily mean it's invalid, but this was the case for your `is_valid_token` function. Even if you would have changed its name, creating a utility function for evaluating a simple boolean value is a tough sell.

File: src/aioswitcher/device/tools.py (hunk 8 above)
Comment (TomerFi) [reply src/aioswitcher/device/tools.py]: Well... after five days and ten overall comments just for this review, thanks for not arguing! :wink: 

By the way, it looks like this discussion was moved: https://github.com/TomerFi/aioswitcher/pull/766#discussion_r1569583564

File: src/aioswitcher/device/tools.py (hunk 11 above)
Comment (TomerFi) [reply src/aioswitcher/device/tools.py]: Are you referring to this?
```
src/aioswitcher/device/tools.py:194: error: Argument 1 to "bytes" has incompatible type "str | None"; expected "str"  [arg-type]
```

File: src/aioswitcher/device/tools.py (hunk 7 above)
Comment (TomerFi) [reply src/aioswitcher/device/tools.py]: Are you referring to this?

```
src/aioswitcher/device/tools.py:194: error: Argument 1 to "bytes" has incompatible type "str | None"; expected "str"  [arg-type]
```

Because this issue is correct, you're using string instead of bytes.
```python
# change this
b"jzNrAOjc%lpg3pVr5cF!5Le06ZgOdWuJ"
# to this
b'jzNrAOjc%lpg3pVr5cF!5Le06ZgOdWuJ'
```

I do not see any linting errors related to the assignment you mentioned [above](https://github.com/TomerFi/aioswitcher/pull/766#discussion_r1567556953).

TomerFi: By the way, I really dislike return statements occurring out of context. This just screams "error-prone." Please take another look at my example [here](https://github.com/TomerFi/aioswitcher/pull/766#discussion_r1567465804).

```python
# change this
    ...
        token_packet = hexlify(unpadded_decrypted_value).decode()
    except (KeyError, ValueError) as ve:
        raise RuntimeError("convert token to packet was not successful") from ve
    return token_packet

# to this
    ...
        return hexlify(unpadded_decrypted_value).decode()
    except (KeyError, ValueError) as ve:
        raise RuntimeError("convert token to packet was not successful") from ve
```

src/aioswitcher/device/tools.py
thecode: > I think we should discuss this next PR when I will show you the usage of the convert token to token packet..

So please fix it for now and when we see the usage in the next PR we can discuss it

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +169,38 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
...
+
```
thecode:
```suggestion
```

thecode (changes requested):

src/aioswitcher/device/tools.py
blorby: @YogevBokobza  - can you remove the trail newline and let's merge it?

YogevBokobza:
I took a few days off this PR as I am in the middle of moving apartments.
I will get back to this in the coming week.

blorby:
> I took a few days off this PR as I am in the middle of moving apartments. I will get back to this in the coming week.

Mazal Tov and Good luck!

src/aioswitcher/device/tools.py
YogevBokobza: Done.

src/aioswitcher/device/tools.py
YogevBokobza: Did the changes

YogevBokobza:
Still having a linting issue:
src/aioswitcher/device/tools.py:188: error: Argument 1 to "bytes" has incompatible type "str | None"; expected "str"  [arg-type]
I am switching the token parameter from:
token: Union[str, None] = None
to:
token: str

src/aioswitcher/device/tools.py
```diff
@@ -165,6 +169,28 @@ def convert_str_to_devicetype(device_type: str) -> DeviceType:
     return DeviceType.MINI
 
 
+def convert_token_to_packet(token: str) -> Union[str, None]:
```
thecode: I don't see how this method can return `none`

YogevBokobza: fixed. change to return str.

thecode (approved):

src/aioswitcher/device/tools.py
blorby: @TomerFi - can you review and approve merging please 🥺

TomerFi (approved):

TomerFi:
Merging - thank you everyone!

PR #767, opened 2024-05-19
chore: add s11 to discovery with it's added logic

YogevBokobza:
## Description

>In this code we are adding s11 to the discovery part + it's added logic to support that + refactoring get function to support more than 1 runner

## Checklist

## Additional information

> Anything else?

### DIFF
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -133,7 +133,7 @@ class StateMessageParser:
 
     def get_shutter_position(self, index: int) -> int:
         """Return the current position of the shutter 0 <= pos <= 100."""
-        start_index = 152 + (index * 32 if index else 0)
+        start_index = 152 + (index * 32)
         end_index = start_index + 2
         hex_pos = self._hex_response[start_index:end_index].decode()
         return int(hex_pos, 16)
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -142,9 +142,9 @@ class StateMessageParser:
         """Return the current direction of the shutter (UP/DOWN/STOP)."""
         start_index = 156 + (index * 32)
         end_index = start_index + 4
-        hex_direction = self._hex_response[start_index:end_index].decode()
+        hex_dir = self._hex_response[start_index:end_index].decode()
         directions = dict(map(lambda d: (d.value, d), ShutterDirection))
-        return directions[hex_direction]
+        return directions[hex_dir]
 
 
 @dataclass
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -143,7 +143,7 @@ class StateMessageParser:
         start_index = 156 + (index * 32)
         end_index = start_index + 4
         hex_dir = self._hex_response[start_index:end_index].decode()
-        directions = dict(map(lambda d: (d.value, d), ShutterDirection))
+        directions = dict(map(lambda s: (s.value, s), ShutterDirection))
         return directions[hex_dir]
 
 
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -20,12 +20,13 @@ from typing import Set, final
 
 from ..device import (
     DeviceState,
+    DeviceType,
     ShutterDirection,
     ThermostatFanLevel,
     ThermostatMode,
     ThermostatSwing,
 )
-from ..device.tools import seconds_to_iso_time, watts_to_amps
+from ..device.tools import get_shutter_index, seconds_to_iso_time, watts_to_amps
 from ..schedule.parser import SwitcherSchedule, get_schedules
 
 
@@ -130,16 +131,22 @@ class StateMessageParser:
         remote_hex = unhexlify(self._hex_response)
         return remote_hex[84:92].decode().rstrip("\x00")
 
-    def get_shutter_position(self) -> int:
-        """Return the current shutter position."""
-        hex_pos = self._hex_response[152:154].decode()
+    def get_shutter_position(self, index: int) -> int:
+        """Return the current position of the shutter 0 <= pos <= 100."""
+        index -= 1
+        start_index = 152 + (index * 32 if index else 0)
+        end_index = start_index + 2
+        hex_pos = self._hex_response[start_index:end_index].decode()
         return int(hex_pos, 16)
 
-    def get_shutter_direction(self) -> ShutterDirection:
-        """Return the current shutter direction."""
-        hex_dir = self._hex_response[156:160].decode()
-        directions = dict(map(lambda s: (s.value, s), ShutterDirection))
-        return directions[hex_dir]
+    def get_shutter_direction(self, index: int) -> ShutterDirection:
+        """Return the current direction of the shutter (UP/DOWN/STOP)."""
+        index -= 1
+        start_index = 156 + (index * 32 if index else 0)
+        end_index = start_index + 4
+        hex_direction = self._hex_response[start_index:end_index].decode()
+        directions = dict(map(lambda d: (d.value, d), ShutterDirection))
+        return directions[hex_direction]
 
 
 @dataclass
@@ -255,10 +262,13 @@ class SwitcherShutterStateResponse(SwitcherBaseResponse):
 
     position: int = field(init=False)
     direction: ShutterDirection = field(init=False)
+    device_type: DeviceType
+    index: int
 
     def __post_init__(self) -> None:
         """Post initialization of the message."""
         parser = StateMessageParser(self.unparsed_response)
+        index = get_shutter_index(self.device_type, self.index)
 
-        self.direction = parser.get_shutter_direction()
-        self.position = parser.get_shutter_position()
+        self.direction = parser.get_shutter_direction(index)
+        self.position = parser.get_shutter_position(index)
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -133,7 +133,6 @@ class StateMessageParser:
 
     def get_shutter_position(self, index: int) -> int:
         """Return the current position of the shutter 0 <= pos <= 100."""
-        index -= 1
         start_index = 152 + (index * 32 if index else 0)
         end_index = start_index + 2
         hex_pos = self._hex_response[start_index:end_index].decode()
@@ -141,7 +140,6 @@ class StateMessageParser:
 
     def get_shutter_direction(self, index: int) -> ShutterDirection:
         """Return the current direction of the shutter (UP/DOWN/STOP)."""
-        index -= 1
         start_index = 156 + (index * 32 if index else 0)
         end_index = start_index + 4
         hex_direction = self._hex_response[start_index:end_index].decode()
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -132,14 +132,14 @@ class StateMessageParser:
         return remote_hex[84:92].decode().rstrip("\x00")
 
     def get_shutter_position(self, index: int) -> int:
-        """Return the current position of the shutter 0 <= pos <= 100."""
+        """Return the current shutter position."""
         start_index = 152 + (index * 32)
         end_index = start_index + 2
         hex_pos = self._hex_response[start_index:end_index].decode()
         return int(hex_pos, 16)
 
     def get_shutter_direction(self, index: int) -> ShutterDirection:
-        """Return the current direction of the shutter (UP/DOWN/STOP)."""
+        """Return the current shutter direction."""
         start_index = 156 + (index * 32)
         end_index = start_index + 4
         hex_dir = self._hex_response[start_index:end_index].decode()
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -140,7 +140,7 @@ class StateMessageParser:
 
     def get_shutter_direction(self, index: int) -> ShutterDirection:
         """Return the current direction of the shutter (UP/DOWN/STOP)."""
-        start_index = 156 + (index * 32 if index else 0)
+        start_index = 156 + (index * 32)
         end_index = start_index + 4
         hex_direction = self._hex_response[start_index:end_index].decode()
         directions = dict(map(lambda d: (d.value, d), ShutterDirection))
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -29,7 +29,6 @@ from .device import (
     DeviceCategory,
     DeviceState,
     DeviceType,
-    LightState,
     ShutterDirection,
     SwitcherBase,
     SwitcherPowerPlug,
@@ -170,8 +169,10 @@ def _parse_device_from_datagram(
                     device_type.token_needed,
                     parser.get_shutter_position(get_shutter_index(device_type, 0)),
                     parser.get_shutter_direction(get_shutter_index(device_type, 0)),
-                    parser.get_light_state(get_light_index(device_type, 0)),
-                    parser.get_light_state(get_light_index(device_type, 1)),
+                    [
+                        parser.get_light_state(get_light_index(device_type, 0)),
+                        parser.get_light_state(get_light_index(device_type, 1)),
+                    ],
                 )
             )
 
@@ -428,14 +429,16 @@ class DatagramParser:
         directions = dict(map(lambda d: (d.value, d), ShutterDirection))
         return directions[hex_direction]
 
-    def get_light_state(self, index: int) -> LightState:
+    def get_light_state(self, index: int) -> DeviceState:
         """Extract the light state from the broadcast message."""
         start_index = 135 + (index * 16)
         end_index = start_index + 2
         hex_pos = hexlify(self.message[start_index:end_index]).decode()
         hex_device_state = hex_pos[0:2]
         return (
-            LightState.ON if hex_device_state == LightState.ON.value else LightState.OFF
+            DeviceState.ON
+            if hex_device_state == DeviceState.ON.value
+            else DeviceState.OFF
         )
 
     # Switcher Breeze methods
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -415,14 +415,14 @@ class DatagramParser:
 
     def get_shutter_position(self, index: int) -> int:
         """Return the current position of the shutter 0 <= pos <= 100."""
-        start_index = 135 if index == 0 else 135 + (index * 16)
+        start_index = 135 + (index * 16)
         end_index = start_index + 2
         hex_pos = hexlify(self.message[start_index:end_index]).decode()
         return int(hex_pos[2:4]) + int(hex_pos[0:2], 16)
 
     def get_shutter_direction(self, index: int) -> ShutterDirection:
         """Return the current direction of the shutter (UP/DOWN/STOP)."""
-        start_index = 137 if index == 0 else 137 + (index * 16)
+        start_index = 137 + (index * 16)
         end_index = start_index + 2
         hex_direction = hexlify(self.message[start_index:end_index]).decode()
         directions = dict(map(lambda d: (d.value, d), ShutterDirection))
@@ -430,7 +430,7 @@ class DatagramParser:
 
     def get_light_state(self, index: int) -> LightState:
         """Extract the light state from the broadcast message."""
-        start_index = 135 if index == 0 else 135 + (index * 16)
+        start_index = 135 + (index * 16)
         end_index = start_index + 2
         hex_pos = hexlify(self.message[start_index:end_index]).decode()
         hex_device_state = hex_pos[0:2]
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -29,17 +29,24 @@ from .device import (
     DeviceCategory,
     DeviceState,
     DeviceType,
+    LightState,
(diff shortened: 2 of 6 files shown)


TheSpringBoy:
Hi, is it possible to add support to the light switches as well? Since you already dealt with the lights here..

YogevBokobza:
> Hi, is it possible to add support to the light switches as well? Since you already dealt with the lights here..

It is not related to this PR..
I rather not to add more parts and I am trying to have the minimum code every time so we can make a progress faster..

src/aioswitcher/api/messages.py
```diff
@@ -130,16 +131,22 @@ def get_thermostat_remote_id(self) -> str:
         remote_hex = unhexlify(self._hex_response)
         return remote_hex[84:92].decode().rstrip("\x00")
 
-    def get_shutter_position(self) -> int:
-        """Return the current shutter position."""
-        hex_pos = self._hex_response[152:154].decode()
+    def get_shutter_position(self, index: int) -> int:
+        """Return the current position of the shutter 0 <= pos <= 100."""
+        index -= 1
```
thecode: I see that the `index -= 1` is repeating in multiple places here, wouldn't it be better to use zero based indexing? after all this is something internal and we programmers preferer counting from zero.

YogevBokobza: Done.

thecode: > Done.

Where exactly is it done?

YogevBokobza: https://github.com/TomerFi/aioswitcher/pull/767/files#diff-76fda1abe206ffc78578e59e694e42cf998f9158ec5427d72e3f42d8db31c7a4R226
Here I can see the change..

YogevBokobza: https://github.com/TomerFi/aioswitcher/pull/767/commits/630a921d800c2fdd81306cf031b9f4e5397b104c

YogevBokobza: Oh this is something else.. let me see

YogevBokobza: In this case, I can't change this part..
This is part of the logic..
If the index is 1 then it should be 0 if this is 2 then it should be 1 and so on..
This is how this works..

thecode: Everywhere in this PR you reduce 1 from the index (about 6 occasion), than the only place you can get the index is a method you implanted which can return 1, 3 or 0 (which was returning -1 that I also don't understand why it is now changed to 0 and what is it used for).
Why can't this method return 0 or 2 so you don't have to reduce it by one later everywhere?

> This is how this works..

That is not how PR review works, there is a method that return -1 (changed to 0 without any explanation why), 3 or 1 without any documentation what it means, you have no plan to explain why you wrote it like that, I don't mind we resolve this conversation and I let someone else review it.

YogevBokobza: I will try to explain..
The changes here: https://github.com/TomerFi/aioswitcher/pull/767/commits/630a921d800c2fdd81306cf031b9f4e5397b104c
I changed based on your request as here it does not matter if this is -1 or 0 because if the conditions are not applied it does not matter what is returning.

The other index -= 1 are parts of this logic to return the correct packet and it should remain..

thecode: I think we are having a broken discussion. I didn't ask to return 0 instead of -1 from `get_shutter_index`, if this condition can't be met you should either return `None` or raise.

I asked that we use zero based index instead of having to do `index -= 1` on all of the methods here. I don't see this breaking any method as instead of passing 3 to a method and let it subtract 1 to get 2, you can just pass 2.
I fully understand from the code here how the packets are generated and I don't understand your comment:
> The other index -= 1 are parts of this logic to return the correct packet and it should remain..

So please explain it better if I am missing something.

thecode:
```suggestion
```

thecode (changes requested):
Please use zero based index. Fix methods that are used to get the index to count from zero so the code will have the same result.

src/aioswitcher/api/messages.py
YogevBokobza: Honestly I don't agree with this change.
You have in your house zero and one shutters?
No you have one and two.
I want go and get shutter number one and shutter number 2.
Using 0 and 1 make no sense here..

Using 0 as a starting point makes sense for indexing that outside of human logic mainly for array, but here you are talking about the shutter number you want to control.
If you buy one Switcher Runner you can control shutter with id 1 which is one shutter you have.

thecode: Why does it have to be related to what the user see? User can still see the first shutter as 1, it would probably mean that we will have one or two places that we have to add one when we display it to the user, but it doesn't make sense that all low level APIs will just reduce 1 from the input, they are not exposed to the user

TomerFi: I agree with @thecode on this. Starting an indexed list from 1 instead of 0 might be in line with the user's perspective, but this is error-prone. A contributor might add some logic and look for the 0 index member. 

Also, changing the value of an argument from the function's signature is a bad practice unless the function is meant to be used recursively. But this is irrelevant if you switch to zero-based indexing and dump the `index -= 1` statement.

src/aioswitcher/device/tools.py
```diff
@@ -211,3 +213,24 @@ def validate_token(username: str, token: str) -> bool:
     else:
         logger.debug("request failed with status code: %s", response.status_code)
     return is_token_valid
+
+
+def get_shutter_index(device_type: DeviceType, device_num: int) -> int:
```
TomerFi: This is very weird and requires some explanation in the code docs.

src/aioswitcher/device/tools.py
```diff
@@ -211,3 +213,24 @@ def validate_token(username: str, token: str) -> bool:
...
+        if device_num == 1:
+            return 1
+    if device_type == DeviceType.RUNNER_S11:
+        if device_num == 1:
+            return 3
+    return 0
+
+
+def get_light_index(device_type: DeviceType, device_num: int) -> int:
+    """Return the currect light index (based of device type)."""
```
TomerFi: Please explain this.

src/aioswitcher/device/__init__.py
```diff
@@ -121,6 +121,31 @@ def value(self) -> str:
         return self._value  # type: ignore
 
 
+@final
+class LightState(Enum):
```
TomerFi: Is this identical to DeviceState?

TomerFi (changes requested):

src/aioswitcher/api/messages.py
YogevBokobza: Change has been done..

YogevBokobza: Change has been done + added testes

src/aioswitcher/device/tools.py
YogevBokobza: Added more docs to the functions description.

src/aioswitcher/device/tools.py
YogevBokobza: I will explain.
get_shutter_index is used to get the correct position to look for in the packet for getting the shutter's direction/position.
get_light_index is used to get the correct position to look for in the packet for getting the light status if it's on or off.

The main idea of the position is like this:

- RUNNER/RUNNER MINI:
The light index does not exist
The shutter index is 0
- Runner S11:
The first light index is 0
The second light index is 1
The shutter index is 2
- Runner S12:
The light index is 0
The first shutter index is 1
The second shutter index is 2

So basically, if the device has lights, it comes before the shutter.

I could remove get_shutter_index and get_light_index altogether and put these hardcoded values inside the bridge.py file.
I just figured it cleaner to look on this part like this.

src/aioswitcher/device/__init__.py
YogevBokobza: Yes and no..
DeviceState is meant to show the device's state..
But let's say Runner S11 which Runner is the device but it has 2 different modules in it which is 2 lights.
I think it makes more sense to have LightState for each of them because they are not devices inside the device..

src/aioswitcher/api/messages.py
```diff
@@ -130,16 +131,20 @@ def get_thermostat_remote_id(self) -> str:
         remote_hex = unhexlify(self._hex_response)
         return remote_hex[84:92].decode().rstrip("\x00")
 
-    def get_shutter_position(self) -> int:
-        """Return the current shutter position."""
-        hex_pos = self._hex_response[152:154].decode()
+    def get_shutter_position(self, index: int) -> int:
+        """Return the current position of the shutter 0 <= pos <= 100."""
+        start_index = 152 + (index * 32 if index else 0)
```
thecode: The `if index` has no effect
```suggestion
        start_index = 152 + (index * 32)
```

src/aioswitcher/api/messages.py
```diff
@@ -130,16 +131,20 @@ def get_thermostat_remote_id(self) -> str:
...
         return int(hex_pos, 16)
 
-    def get_shutter_direction(self) -> ShutterDirection:
-        """Return the current shutter direction."""
-        hex_dir = self._hex_response[156:160].decode()
-        directions = dict(map(lambda s: (s.value, s), ShutterDirection))
-        return directions[hex_dir]
+    def get_shutter_direction(self, index: int) -> ShutterDirection:
+        """Return the current direction of the shutter (UP/DOWN/STOP)."""
+        start_index = 156 + (index * 32 if index else 0)
```
thecode: The `if index` has no effect
```suggestion
        start_index = 156 + (index * 32)
```

src/aioswitcher/bridge.py
```diff
@@ -380,19 +411,33 @@ def get_device_type(self) -> DeviceType:
...
         return devices[hex_model]
 
-    # Switcher Runner and Runner Mini methods
+    # Switcher Runners methods
 
-    def get_shutter_position(self) -> int:
+    def get_shutter_position(self, index: int) -> int:
         """Return the current position of the shutter 0 <= pos <= 100."""
-        hex_pos = hexlify(self.message[135:137]).decode()
+        start_index = 135 if index == 0 else 135 + (index * 16)
```
thecode: Same, `if index` is not needed

src/aioswitcher/bridge.py
```diff
@@ -380,19 +411,33 @@ def get_device_type(self) -> DeviceType:
...
+        end_index = start_index + 2
+        hex_pos = hexlify(self.message[start_index:end_index]).decode()
         return int(hex_pos[2:4]) + int(hex_pos[0:2], 16)
 
-    def get_shutter_direction(self) -> ShutterDirection:
+    def get_shutter_direction(self, index: int) -> ShutterDirection:
         """Return the current direction of the shutter (UP/DOWN/STOP)."""
-        hex_direction = hexlify(self.message[137:139]).decode()
+        start_index = 137 if index == 0 else 137 + (index * 16)
```
thecode: Same, `if index` is not needed

src/aioswitcher/bridge.py
```diff
@@ -380,19 +411,33 @@ def get_device_type(self) -> DeviceType:
...
+        end_index = start_index + 2
+        hex_direction = hexlify(self.message[start_index:end_index]).decode()
         directions = dict(map(lambda d: (d.value, d), ShutterDirection))
         return directions[hex_direction]
 
+    def get_light_state(self, index: int) -> LightState:
+        """Extract the light state from the broadcast message."""
+        start_index = 135 if index == 0 else 135 + (index * 16)
```
thecode: Same, `if index` is not needed

thecode (changes requested):

src/aioswitcher/bridge.py
YogevBokobza: Fixed..

src/aioswitcher/bridge.py
YogevBokobza: Fixed..

src/aioswitcher/bridge.py
YogevBokobza: Fixed..

src/aioswitcher/api/messages.py
```diff
@@ -130,16 +131,20 @@ def get_thermostat_remote_id(self) -> str:
...
-        hex_dir = self._hex_response[156:160].decode()
-        directions = dict(map(lambda s: (s.value, s), ShutterDirection))
-        return directions[hex_dir]
+    def get_shutter_direction(self, index: int) -> ShutterDirection:
+        """Return the current direction of the shutter (UP/DOWN/STOP)."""
+        start_index = 156 + (index * 32)
+        end_index = start_index + 4
+        hex_direction = self._hex_response[start_index:end_index].decode()
+        directions = dict(map(lambda d: (d.value, d), ShutterDirection))
+        return directions[hex_direction]
```
thecode: A general comment not a request for changes, the following 3 lines renaming variables without any need, please try to avoid it in feature PRs and separate such changes if they are needed to code quality PRs. 
In this PR 2 lines would not even show on the review if the name `hex_dir` was kept as it is.

src/aioswitcher/device/tools.py
```diff
@@ -211,3 +213,32 @@ def validate_token(username: str, token: str) -> bool:
...
     return is_token_valid
+
+
+def get_shutter_index(device_type: DeviceType, device_num: int) -> int:
+    """Return the correct shutter index.
+
+    Used in retriving the shutter position/direction from the packet
+    (based of device type and device number).
+    """
+    if device_type == DeviceType.RUNNER or device_type == DeviceType.RUNNER_MINI:
```
thecode: I can only assume this is a preparation for future devices not `S11` but for now it doesn't make sense to check `device_num` in each conditon.
```python
if device_num != 0:
    raise ValueError("Invalid device number")

if device_type in (DeviceType.RUNNER, DeviceType.RUNNER_MINI):
   return 0
elif device_type == DeviceType.RUNNER_S11:
  return 2

raise ValueError("only shutters are allowed")
```

src/aioswitcher/api/messages.py
YogevBokobza: Should I revert to hex_dir or leave it like this?

src/aioswitcher/device/tools.py
thecode: I think even after the update it is not clear.
Please add a small part that explain the device configuration. something like:
S11 has two lights circuits & one shutter circuits.
Lights circuits are numbered 0 & 1, shutter circuit is 0.

I also think the name `device_num` is confusing as we use device in multiple places here. I don't know what would be the best option but maybe `circuit_number` or `sub_device_number`
I usually try to avoid short names in external params, I prefer it will be clear what it does vs a shorter code.

src/aioswitcher/device/__init__.py
```diff
@@ -321,6 +346,23 @@ class SwitcherShutterBase(ABC):
...
+        position: the current position of the shutter (integer percentage).
+        direction: the current direction of the shutter.
+        light: the current light state.
+        light2: the current second light state.
+    """
+
+    position: int
+    direction: ShutterDirection
+    light: LightState
+    light2: LightState
```
thecode: Maybe it would be better to make this an array?

src/aioswitcher/api/messages.py
thecode: I think it would be better to revert to `hex_dir` just to make it simpler to understand the changes in this message.

src/aioswitcher/device/__init__.py
YogevBokobza: It is possible but I think it makes things more complicated and not that much readable if you consider the maximum number of lights possible in a device is 3 using the product: Switcher Light SL03

src/aioswitcher/device/tools.py
YogevBokobza: Regarding:
`
S11 has two lights circuits & one shutter circuits.
Lights circuits are numbered 0 & 1, shutter circuit is 0.
`
Where is the best place to write this?

YogevBokobza: @thecode I put this information next to those functions.. I am not sure where to put it if not here..
Also changed to circuit_number

src/aioswitcher/device/tools.py
YogevBokobza: Changed to the code suggested.

src/aioswitcher/device/__init__.py
```diff
@@ -376,3 +418,15 @@ def __post_init__(self) -> None:
...
         return super().__post_init__()
+
+
+@final
+@dataclass
+class SwitcherSingleShutterDualLight(SwitcherSingleShutterDualLightBase, SwitcherBase):
+    """Implementation of the Switcher Shutter with dual light device."""
+
+    def __post_init__(self) -> None:
+        """Post initialization validate device type category as SINGLE_SHUTTER_DUAL_LIGHT."""  # noqa: E501
```
thecode: Please break into two lines instead of adding #noqa

src/aioswitcher/device/tools.py
thecode: From my point it looks clear now

src/aioswitcher/device/__init__.py
YogevBokobza: I tried to change to
"""Post initialization validate device type category
        as SINGLE_SHUTTER_DUAL_LIGHT.
        """

but got this lint errors:
src/aioswitcher\device\__init__.py:429:1: D205 1 blank line required between summary line and description
src/aioswitcher\device\__init__.py:429:1: D400 First line should end with a period

thecode: That is correct, when documenting a function you should have a short description on the first line, than a black line and after you can multiple lines, something like this should work:
```
"""Post initialization

Validate device type category as SINGLE_SHUTTER_DUAL_LIGHT."""
```

src/aioswitcher/device/__init__.py
thecode: So when you add the 3rd light it will be `light3`? wouldn't it be better to have an array, than when you add additional light to the array it will just expand by one?

I leave @TomerFi to suggest what he thinks would be better here

YogevBokobza: I think an array is unneeded here and can make even more confusions, but let's leave it for @TomerFi

src/aioswitcher/device/__init__.py
YogevBokobza: src/aioswitcher\device\__init__.py:429:1: D209 Multi-line docstring closing quotes should be on a separate line
src/aioswitcher\device\__init__.py:429:1: D400 First line should end with a period

Still issue..

thecode: Well I just gave an example, yes first line should end with a period and you need to move the 3 """ to a new line.
It seems strange you come back here with this, wouldn't it be simple just to fix it the errors? they are clear.
To be honest I never face these errors since I am using linters which reformat the text as I write it.

YogevBokobza: Done..

src/aioswitcher/device/__init__.py
TomerFi: IDK... It feels weird two identical classes just for their names.
@thecode What do you think?

src/aioswitcher/device/__init__.py
TomerFi: What is the chance we'll need to add another light slot? If we add more, I agree with @thecode: Let's switch to an array. But if it's a long shot, we can stick to vars, maybe change the name to "lightSlot1" and "lightSlot2" or something like that because light+light2 feels like a patch.

Another option might be to have a dict with a string key ("light_slot1", "light_slot2") and a State instance as a value.

thecode: We will need another slot in the future see https://github.com/TomerFi/aioswitcher/pull/767#discussion_r1615241270, Switcher Light SL03 has 3 lights.

src/aioswitcher/device/__init__.py
thecode: I think we can start without it and if we face a problem we can add it later.

src/aioswitcher/device/__init__.py
TomerFi: Gotcha.
Let's do an array. it feels right.

src/aioswitcher/device/__init__.py
YogevBokobza: Removed and changed to DeviceState..

src/aioswitcher/device/__init__.py
YogevBokobza: I changed that to an array.

thecode (dismissed):

src/aioswitcher/device/tools.py
```diff
@@ -211,3 +213,43 @@ def validate_token(username: str, token: str) -> bool:
...
+    """Return the correct light index.
+
+    Used in retriving the light on/off status from the packet
+    (based of device type and circuit number).
+    """
+    if device_type == DeviceType.RUNNER_S11:
+        if circuit_number == 0:
+            return 0
+        elif circuit_number == 1:
+            return 1
```
TomerFi:
```python
if circuit_number in [0, 1]:
    return circuit_number
```

TomerFi (changes requested):

src/aioswitcher/device/tools.py
YogevBokobza: Done..

thecode (approved):

TomerFi (approved):

TomerFi:
@YogevBokobza @thecode - Thank you very much!
I'm merging this.

Also, we need to rebase runner-support-fixed from the main.

PR #777, opened 2024-06-14
chore: add s11 control device

YogevBokobza:
## Description

## Checklist

## Additional information

> Anything else?


src/aioswitcher/api/__init__.py
```diff
@@ -103,6 +108,10 @@ def __init__(
         self._device_key = device_key
         self._port = port
         self._connected = False
+        if self._device_type.token_needed and bool(token):
+            self._token = convert_token_to_packet(str(token))
+        else:
+            self._token = ""
```
TomerFi: Why not `None`?

src/aioswitcher/api/__init__.py
```diff
@@ -162,20 +171,33 @@ async def _login(self) -> Tuple[str, SwitcherLoginResponse]:
...
+            packet = packets.LOGIN_PACKET_TYPE2.format(timestamp, self._device_id)
         else:
             packet = packets.LOGIN_PACKET_TYPE1.format(timestamp, self._device_key)
         signed_packet = sign_packet_with_crc_key(packet)
 
         logger.debug("sending a login packet")
         self._writer.write(unhexlify(signed_packet))
         response = await self._reader.read(1024)
+
+        if self._device_type and self._device_type.token_needed and bool(self._token):
```
TomerFi: In these 20 lines, we evaluate `self._device_type` three times. Is an empty `self._device_type` even feasible? Isn't it an _Enum_?

docs/scripts.md
```diff
@@ -84,6 +84,8 @@ subcommands:
     stop_shutter        stop shutter
     turn_off            turn off the device
     turn_on             turn on the device
+    turn_off_light      turn off light
+    turn_on_light       turn on light
```
TomerFi, line 88: I think `light_on` and `light_off` might better distinguish between `turn_off` and `turn_on`.

YogevBokobza:
@TomerFi @thecode
Please try to go over this PR when you can..

src/aioswitcher/api/__init__.py
```diff
@@ -103,6 +108,10 @@ def __init__(
         self._device_key = device_key
         self._port = port
         self._connected = False
+        if self._device_type.token_needed and bool(token):
```
thecode: 1. It doesn't make sense to convert a type of `str` or  `None` to a Boolean, why not just checking if it is not `None`?
2. What happens if token is needed but is not provided?
3. The value already have a default value of `None` but you later override it to empty string `""`, either work with it as `None` (IMO cleanest way) or set a default value of `""` (e.g. `token: str = ""`)

src/aioswitcher/api/__init__.py
```diff
@@ -702,13 +758,17 @@ async def _control_breeze_swing_device(
...
         """Use for stopping the shutter.
 
+        Args:
+            index: which runner to stop position, default to 0.
+
         Returns:
             An instance of ``SwitcherBaseResponse``.
 
         """
+        index = get_shutter_index(self._device_type, index) + 1
```
thecode: Don't override the function input parameters, if you want to modify the index do it on a new variable

src/aioswitcher/api/__init__.py
```diff
@@ -719,9 +779,21 @@ async def stop_shutter(self) -> SwitcherBaseResponse:
             "logged in session_id=%s, timestamp=%s", login_resp.session_id, timestamp
         )
 
-        packet = packets.RUNNER_STOP_COMMAND.format(
-            login_resp.session_id, timestamp, self._device_id
-        )
+        if self._device_type and self._device_type.token_needed and bool(self._token):
```
thecode: This check is redundant, when you create the class you already checked for the conditions for the token, here just use the token if it is already set or skip the token if it is not set

src/aioswitcher/api/__init__.py
```diff
@@ -719,9 +779,21 @@ async def stop_shutter(self) -> SwitcherBaseResponse:
...
+            command = "0000"
+            hex_pos = f"0{index}{command}" if index else command
```
thecode: `if index` doesn't make sense since you set above `index+1` so index can't be zero

tests/test_api_tcp_client.py
```diff
@@ -49,10 +49,13 @@
 
 device_type_api1 = DeviceType.TOUCH
 device_type_api2 = DeviceType.RUNNER
-device_index = 2
+device_type_token_api2 = DeviceType.RUNNER_S11
+device_index = 0
```
thecode, line 53: Please add a test for index != 0 since the packet is different for index which is not 0

thecode (changes requested):

src/aioswitcher/api/__init__.py
YogevBokobza: 1. Because of this reason:
- We don't need to convert the token to a packet if the device does not require a token.. This is to prevent selecting a device that does not need a token and providing a token.
2. It should send the API calls to the token-based packets with an empty string token and will result in no action to the device.
3. Because self._token is later used for the packet sent to the device it should be at least an empty string.. I could split that into 2 variables: self._token and self._token_packet if this makes more sense, I thought it just needed..

src/aioswitcher/api/__init__.py
YogevBokobza: Will do

tests/test_api_tcp_client.py
YogevBokobza: It is done.

src/aioswitcher/api/__init__.py
YogevBokobza: It is done.

src/aioswitcher/api/__init__.py
YogevBokobza: it is done.

src/aioswitcher/api/__init__.py
YogevBokobza: it is done.

scripts/control_device.py
```diff
@@ -52,7 +52,14 @@
...
 
 python control_device.py stop_shutter -c "Switcher Runner" -d f2239a -i "192.168.50.98"\n
+python control_device.py stop_shutter -c "Switcher Runner S11" -k "zvVvd7JxtN7CgvkD1Psujw==" -d f2239a -i "192.168.50.98"\n
 python control_device.py set_shutter_position -c "Switcher Runner" -d f2239a -i "192.168.50.98" -p 50\n
+python control_device.py set_shutter_position -c "Switcher Runner S11" -k "zvVvd7JxtN7CgvkD1Psujw==" -d f2239a -i "192.168.50.98" -p 50\n
+
+python control_device.py turn_on_light -c "Switcher Runner S11" -k "zvVvd7JxtN7CgvkD1Psujw==" -d ab1c2d -i "111.222.11.22" -x 0\n
+python control_device.py turn_on_light -c "Switcher Runner S11" -k "zvVvd7JxtN7CgvkD1Psujw==" -d ab1c2d -i "111.222.11.22" -x 1\n
+python control_device.py turn_off_light -c "Switcher Runner S11" -k "zvVvd7JxtN7CgvkD1Psujw==" -d ab1c2d -i "111.222.11.22" -x 0\n
+python control_device.py turn_off_light -c "Switcher Runner S11" -k "zvVvd7JxtN7CgvkD1Psujw==" -d ab1c2d -i "111.222.11.22" -x 1\n
```
TomerFi, line 62: I'm just making sure the token in these examples isn't a real value dropped here by mistake. Feel free to resolve this if it isn't.

scripts/control_device.py
```diff
@@ -273,11 +287,27 @@
...
     help="Shutter position percentage",
 )
+set_shutter_position_parser.add_argument(
+    "-x",
+    "--index",
+    required=False,
+    type=int,
+    default=0,
+    help="select shutter number id to operate",
+)
```
TomerFi: What does the shutter number ID stand for?

src/aioswitcher/api/__init__.py
TomerFi: > It doesn't make sense to convert a type of str or None to a Boolean, why not just checking if it is not None?

I'm ok with _bool(token)_. The [bool](https://www.w3schools.com/python/ref_func_bool.asp) function in Python is used more to evaluate expressions than convert. It's commonly used and will return false for `None` and `""`.

The other two questions still stand.

src/aioswitcher/api/__init__.py
```diff
@@ -162,20 +171,33 @@ async def _login(self) -> Tuple[str, SwitcherLoginResponse]:
 
         """
         timestamp = current_timestamp_to_hexadecimal()
-        if (
+        if self._token != "":
```
TomerFi: Why not bool(self._token)?

src/aioswitcher/api/__init__.py
```diff
@@ -162,20 +171,33 @@ async def _login(self) -> Tuple[str, SwitcherLoginResponse]:
...
         response = await self._reader.read(1024)
+
+        if self._token != "":
+            packet = packets.LOGIN2_TOKEN_PACKET_TYPE2.format(
+                self._device_id, timestamp, self._token
+            )
+            signed_packet = sign_packet_with_crc_key(packet)
+            logger.debug("sending a login2 packet")
+            self._writer.write(unhexlify(signed_packet))
+            response = await self._reader.read(1024)
```
TomerFi: 1. Why not bool(self._token)?
2. Should this section be merged with the section in row 174 instead of evaluating for the same expression twice?

src/aioswitcher/api/__init__.py
```diff
@@ -556,11 +602,21 @@ class SwitcherType2Api(SwitcherApi):
     """
 
     def __init__(
-        self, device_type: DeviceType, ip_address: str, device_id: str, device_key: str
+        self,
+        device_type: DeviceType,
+        ip_address: str,
+        device_id: str,
+        device_key: str,
+        token: Union[str, None] = None,
```
TomerFi, line 610: If the token default value is None and it's the last argument in the signature, why do all examples not require a token to pass an empty string?

src/aioswitcher/api/__init__.py
```diff
@@ -719,9 +779,21 @@ async def stop_shutter(self) -> SwitcherBaseResponse:
             "logged in session_id=%s, timestamp=%s", login_resp.session_id, timestamp
         )
 
-        packet = packets.RUNNER_STOP_COMMAND.format(
-            login_resp.session_id, timestamp, self._device_id
-        )
+        if self._token != "":
```
TomerFi: Why not bool(self._token)?

src/aioswitcher/api/__init__.py
```diff
@@ -732,16 +804,20 @@ async def stop_shutter(self) -> SwitcherBaseResponse:
...
 
         Args:
             position: the position to set the device to, default to 0.
+            index: which runner to set position, default to 0.
 
         Returns:
             An instance of ``SwitcherBaseResponse``.
 
         """
+        index_packet = get_shutter_index(self._device_type, index) + 1
```
TomerFi: Please add a comment explaining the `+1` part.

src/aioswitcher/api/__init__.py
```diff
@@ -754,9 +830,20 @@ async def set_position(self, position: int = 0) -> SwitcherBaseResponse:
             "logged in session_id=%s, timestamp=%s", login_resp.session_id, timestamp
         )
 
-        packet = packets.RUNNER_SET_POSITION.format(
-            login_resp.session_id, timestamp, self._device_id, hex_pos
-        )
+        if self._token != "":
```
TomerFi: Why not bool(self._token)?

src/aioswitcher/api/__init__.py
```diff
@@ -828,3 +915,50 @@ async def get_shutter_state(self, index: int = 0) -> SwitcherShutterStateRespons
...
+        timestamp, login_resp = await self._login()
+        if not login_resp.successful:
+            logger.error("Failed to log into device with id %s", self._device_id)
+            raise RuntimeError("login request was not successful")
+
+        logger.debug(
+            "logged in session_id=%s, timestamp=%s", login_resp.session_id, timestamp
+        )
+
+        if self._token != "":
```
TomerFi: Why not bool(self._token)?

src/aioswitcher/api/__init__.py
```diff
@@ -828,3 +915,50 @@ async def get_shutter_state(self, index: int = 0) -> SwitcherShutterStateRespons
...
+
+        Args:
+            command: use the ``aioswitcher.api.DeviceState`` enum.
+            index: which light to turn on/off, default to 0.
+
+        Returns:
+            An instance of ``SwitcherBaseResponse``.
+
+        """
+        index_packet = get_light_index(self._device_type, index) + 1
```
TomerFi: Please add a comment explaining the +1 part.

src/aioswitcher/api/packets.py
```diff
@@ -25,13 +25,59 @@
...
+STOP_SHUTTER_PRECOMMAND = "3702"
+SET_POSITION_PRECOMMAND = "3701"
+SET_LIGHT_PRECOMMAND = "370a"
+
+# login packet for Type1 devices
+# format value just timestamp (initial session id is P_SESSION)
 LOGIN_PACKET_TYPE1 = (
-    "fef052000232a10000000000" + REQUEST_FORMAT_TYPE1[2:] + "{}" + PAD_72_ZEROS + "00"
+    "fef052000232a100"
+    + P_SESSION
```
TomerFi, line 39: Are we sure this won't break?

src/aioswitcher/api/packets.py
```diff
@@ -25,13 +25,59 @@
...
+    "fef053000305a100"
+    + P_SESSION
+    + "f50301000000"
+    + "{}"
+    + "000000"
+    + "{}"
+    + "00000000000000000000f0fe"
+    + "0000"
+    + "{}"
+    + "000000000000000000000000000000000000000000000000000000000000000001"
```
TomerFi, line 80: Are we sure this won't break?

TomerFi (changes requested):

ido-re:
@thecode @TomerFi -we'll be happy to get the Switcher light integrations to HASS :)

scripts/control_device.py
YogevBokobza: Of course, it isn't real value.. Just an example.
Resolving..

scripts/control_device.py
YogevBokobza: Basically, that is the circuit number I will update the help string

docs/scripts.md
YogevBokobza: I wanted to keep the same convention here.. Do you really think this needs a change?

src/aioswitcher/api/__init__.py
YogevBokobza: There is only the 1 now in line 179.. this code is outdated..

src/aioswitcher/api/__init__.py
YogevBokobza: I explained that [here](https://github.com/TomerFi/aioswitcher/pull/777#discussion_r1656693738)

src/aioswitcher/api/packets.py
YogevBokobza: Yes, this is tested.. It is the same end results but with formatted differently to make more sense.

src/aioswitcher/api/packets.py
YogevBokobza: Yes, this is tested.. It is the same end results but with formatted differently to make more sense.

scripts/control_device.py
YogevBokobza: it is done.

src/aioswitcher/api/__init__.py
YogevBokobza: it is done.

src/aioswitcher/api/__init__.py
YogevBokobza: it is done.

src/aioswitcher/api/__init__.py
YogevBokobza: it is done.

src/aioswitcher/api/__init__.py
YogevBokobza: it is done.

src/aioswitcher/api/__init__.py
YogevBokobza: 1. It is done
2. No because this is a different packet sent and this should come after. so basically there are 2 calls one after the other only if bool(self._token)

src/aioswitcher/api/__init__.py
YogevBokobza: It is done.

src/aioswitcher/api/__init__.py
YogevBokobza: It is done.

src/aioswitcher/api/__init__.py
YogevBokobza: Not sure I got what you mean..

src/aioswitcher/api/__init__.py
TomerFi: Do you mean you use the token even if it is not required, so you must keep it as a string?

src/aioswitcher/api/__init__.py
TomerFi: Thanks, that's better. But I wonder if an Enum can be empty.

docs/scripts.md
TomerFi: IMO, it's more confusing this way.
@thecode What do you think?

scripts/control_device.py
TomerFi: You've changed "select shutter number ID to operate" to "select circuit number to operate." what's the difference?
Maybe it's just my lack of familiarity with the new Switcher devices; perhaps explain in a comment what "shutter number id" / "circuit number" means to help me understand.

src/aioswitcher/api/__init__.py
TomerFi: In the README, you added the following example use cases:
```
asyncio.run(control_runner(DeviceType.RUNNER, "111.222.11.22", "ab1c2d", "00", ""))
asyncio.run(control_runner(DeviceType.RUNNER_MINI, "111.222.11.22", "ab1c2d", "00", ""))
asyncio.run(control_runner(DeviceType.RUNNER_S11, "111.222.11.22", "ab1c2d", "00", "zvVvd7JxtN7CgvkD1Psujw=="))
```

We are missing the concept of the default value. If you change the following two statements, the "token" value will be None, per the default value assigned. Passing an empty string is unnecessary when no token is required.

```
asyncio.run(control_runner(DeviceType.RUNNER, "111.222.11.22", "ab1c2d", "00"))
asyncio.run(control_runner(DeviceType.RUNNER_MINI, "111.222.11.22", "ab1c2d", "00"))
```

src/aioswitcher/api/__init__.py
TomerFi: I'm sorry, I'm not too familiar with Switcher shutters, so this still makes no sense to me.
But if Shay is okay with this, I'm too.

@thecode Are we okay with this +1 and the following comment:
```
# We need to convert the selected circuit number to the actual place in the packet?
```
Do you think this makes sense?

src/aioswitcher/api/packets.py
TomerFi: @thecode I'm Just making sure you're ok with this before resolving.

src/aioswitcher/api/__init__.py
TomerFi: I've checked, and everywhere you use the `self._token,` you first evaluate its value with the `bool(self._token)` prior, which means that `"` and `None` are treated the same in our code. So we can change this to None, and this should work.

src/aioswitcher/api/__init__.py
TomerFi: An example would be better. Consider the following code:

```python
from enum import Enum


class MyEnum(Enum):
    Member1 = 1
    Member2 = 2


class MyClass:
    def __init__(self, member: MyEnum):
        self._member = member

    def member(self) -> MyEnum:
        return self._member


if __name__ == "main":
    print(MyClass(None).member())
```

When you lint this with _MyPy_, you get the following error:

```
error: Argument 1 to "MyClass" has incompatible type "None"; expected "MyEnum"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

So our _MyPy_ CI will fail if a contribution includes passing `None` as the device type.

In our documentation, we instruct users to use _MyEnum_ when instantiating _MyClass_. If they don't adhere, the code should either break or act unexpectedly.

---

Your contribution includes the following:

```python
elif (
            self._device_type
            and self._device_type == DeviceType.BREEZE
            or self._device_type == DeviceType.RUNNER
            or self._device_type == DeviceType.RUNNER_MINI
        ):
```

My question was: Why are you evaluating `self._device_type`? If the device type is `None,` that means the user deliberately instantiated the API object that way.
This also means the next statement `self._device_type == DeviceType.BREEZE` will break the code, which is exactly what I think we should do! Break the code if the device type is `None`.

- Do we have any supported use cases in our code where we don't need to know the device type?
- Do we have a "default" device type?

If the answer to these questions is no, then not only should this statement be removed, but we should also consider verifying the device type at the init function and throwing an error if needed.

@thecode What do you think?

thecode: When we started we had only one type of API (so all devices used the same type), later other types added (so most devices fall into default types) and now most calls are depended on the type of the device, reviewing the state of the implementation now it can easily break if type is not supplied I would just suggest to make it a mandatory.

I suggest to expose only one API init function (in future PR, not this one maybe) and call the different APIs internally by the type of the device, so the user doesn't have to create `SwitcherType1Api`/`SwitcherType2Api` according to the device type, but it will only create a single class - `SwitcherApi` which will internally call the correct logic/messages according to the device type.

docs/scripts.md
thecode: I suggest to keep as implemented by @YogevBokobza since we request a turn on/off action here so it is more clear to keep it in the command.

src/aioswitcher/api/__init__.py
thecode: I don't fully understand the range of the valid indexes for a device since we don't have documentation for how each device works. 

I think we at least need to specify the valid range of this method and what does it mean (circuit number? what is the value for a single circuit device?)

As for the +1, If I understand correctly there are different indexes ranges used by the discovery packet and the API calls, if this is correct might be better to have two methods: `get_discovery_packet_index` and `get_api_packet_index`.

src/aioswitcher/api/__init__.py
thecode: @YogevBokobza so it is not missed, just reminding that above 2 & 3 still need attention

src/aioswitcher/api/__init__.py
TomerFi: That sounds good. I've opened an issue for reference: https://github.com/TomerFi/aioswitcher/issues/778.
Let's remove the `self._device_type` from the evaluation to force a break if no type was selected.

docs/scripts.md
TomerFi: Ok, resolved.

src/aioswitcher/api/__init__.py
YogevBokobza: I thought I already answered lol
Anyway here:
Regarding (2) if a token is needed but is not provided that is what the "if" is for, it should set the token to be empty and the packet will be sent ok but won't do anything in practice.. no control would occur..
Regarding (3) I explained why it cannot stay None... we need to send some string as a token. I said I could set another variable like "self._token_packet" if we don't want to use the same one..

scripts/control_device.py
YogevBokobza: They are the same..
That means which shutter number to operate..
Runner S11 has 1 shutter and 2 lights so the possibilities are shutter 0, light 0 and light 1
Runner S12 has 2 shutters and 1 light so the possibilities are shutter 0, shutter 1 and light 0

I think you are the one that circuit number made more sense so I just aligned that here as well.

src/aioswitcher/api/__init__.py
YogevBokobza: It is done.

src/aioswitcher/api/__init__.py
YogevBokobza: I see now.. that makes sense I guess.
I changed from "" to None in the "else" case.

src/aioswitcher/api/__init__.py
thecode: > packet will be sent ok but won't do anything in practice.. no control would occur.

So what is the point of even checking that? You detected that a token is needed, but it was not provided and you silently ignore it let it just don't work at a later stage. If a token is needed and not provided raise an error, this will also remove the need to constantly check this in later calls.

> explained why it cannot stay None... we need to send some string as a token

This again gets back to (2), if you repeatedly check if a token is set:
https://github.com/TomerFi/aioswitcher/blob/84bc398ec5b55ceb3f7c7e082e25091e207a4409/src/aioswitcher/api/__init__.py#L174
https://github.com/TomerFi/aioswitcher/blob/84bc398ec5b55ceb3f7c7e082e25091e207a4409/src/aioswitcher/api/__init__.py#L193
https://github.com/TomerFi/aioswitcher/blob/84bc398ec5b55ceb3f7c7e082e25091e207a4409/src/aioswitcher/api/__init__.py#L833
https://github.com/TomerFi/aioswitcher/blob/84bc398ec5b55ceb3f7c7e082e25091e207a4409/src/aioswitcher/api/__init__.py#L945

If you evaluate if a token is needed correctly at (2) you will never reach to a situation that you send an empty token, so I would expect the init flow to be:
```python
        self._token = ""
        if self._device_type.token_needed:
            if not token:
                raise RuntimeError("A token is needed but is missing")
            self._token = convert_token_to_packet(str(token))
```
So all of this places can just change from:
`if self._device_type and self._device_type.token_needed and bool(self._token):`
to:
`if self._token:`

YogevBokobza: All of those `if self._device_type and self._device_type.token_needed and bool(self._token): ` lines are gone in the past commits.. please see the latest changes.
I replaced it with the code suggested and added a test to support this Error Raise part.

src/aioswitcher/api/__init__.py
YogevBokobza: > As for the +1, If I understand correctly there are different indexes ranges used by the discovery packet and the API calls, if this is correct might be better to have two methods: get_discovery_packet_index and get_api_packet_index.

I refactored to use 2 functions (discovery and API) for light index and shutter index + added tests.

src/aioswitcher/api/__init__.py
YogevBokobza: It is done..

src/aioswitcher/api/__init__.py
```diff
@@ -162,20 +172,32 @@ async def _login(self) -> Tuple[str, SwitcherLoginResponse]:
...
+            packet = packets.LOGIN_TOKEN_PACKET_TYPE2.format(
+                self._token, timestamp, self._device_id
+            )
+        elif (
+            self._device_type == DeviceType.BREEZE
             or self._device_type == DeviceType.RUNNER
             or self._device_type == DeviceType.RUNNER_MINI
         ):
-            packet = packets.LOGIN2_PACKET_TYPE2.format(timestamp, self._device_id)
+            packet = packets.LOGIN_PACKET_TYPE2.format(timestamp, self._device_id)
```
thecode, line 184:
```suggestion
            packet = packets.LOGIN2_PACKET_TYPE2.format(timestamp, self._device_id)
```
It seems like you change the login for existing devices, are you sure this change is correct and not a typo?

YogevBokobza:
In future PR I will add more tests to fix the coverage percentage I also saw that currently, we are lacking "lights" parts in "message.py" which I will solve with tests next PR.

scripts/control_device.py
TomerFi: Oh, OK, I see.

TomerFi (approved):

src/aioswitcher/api/__init__.py
YogevBokobza: I did not change the login for existing devices. I renamed it to be more clear as requested by @TomerFi.
If you look over the value in this line you will see it evaluates to the same value.
So we have those:

\# first login packet for Type1 devices
LOGIN_PACKET_TYPE1

\# first login packet for Type2 devices
LOGIN_PACKET_TYPE2

\# first login packet for token-based Type2 devices
LOGIN_TOKEN_PACKET_TYPE2

\# second login packet for token-based Type2 devices
LOGIN2_TOKEN_PACKET_TYPE2

LOGIN2_PACKET_TYPE2 is renamed to LOGIN_PACKET_TYPE2.

This is how it looks now:
```
        if bool(self._token):
            packet = packets.LOGIN_TOKEN_PACKET_TYPE2.format(
                self._token, timestamp, self._device_id
            )
        elif (
            self._device_type == DeviceType.BREEZE
            or self._device_type == DeviceType.RUNNER
            or self._device_type == DeviceType.RUNNER_MINI
        ):
            packet = packets.LOGIN_PACKET_TYPE2.format(timestamp, self._device_id)
        else:
            packet = packets.LOGIN_PACKET_TYPE1.format(timestamp, self._device_key)
        signed_packet = sign_packet_with_crc_key(packet)

        logger.debug("sending a login packet")
        self._writer.write(unhexlify(signed_packet))
        response = await self._reader.read(1024)

        if bool(self._token):
            packet = packets.LOGIN2_TOKEN_PACKET_TYPE2.format(
                self._device_id, timestamp, self._token
            )
            signed_packet = sign_packet_with_crc_key(packet)
            logger.debug("sending a login2 packet")
            self._writer.write(unhexlify(signed_packet))
            response = await self._reader.read(1024)
```

So the original code is ok.

thecode (changes requested):
Please answer the unresolved comment

YogevBokobza:
> Please answer the unresolved comment

What unresolved comment? I answered everything..

YogevBokobza:
> Please answer the unresolved comment

You can see now.. Git decided to put my comments in pending..

src/aioswitcher/api/__init__.py
YogevBokobza: @TomerFi Can it be resolved?

src/aioswitcher/api/__init__.py
YogevBokobza: @TomerFi Can it be resolved?

src/aioswitcher/api/__init__.py
YogevBokobza: @TomerFi Can it be resolved?

thecode (approved):
Thanks @YogevBokobza

PR #787, opened 2024-08-06
feat: add switcher s12 support

### LINKED ISSUE #585: Switcher runner S12 support

Hi,

Currently there's support for the single shutter Switcher runner.
Switcher since then started producing a double-shutter single-light device called [Switcher Runner S12](https://switcher.co.il/%D7%9E%D7%95%D7%A6%D7%A8/runner-lights-s12/).
Any chance for adding support to it?

The current API (V2) is unable to nor discover neither control the S12 device.

Yes, add support to a new device

Probably reverse engineer the new API or try to find out the ports open by the new device. I can help if needed.

YogevBokobza:
## Description

> Add Switcher Runner S12 support

**Related issue (if any):** fixes #585

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@thecode @TomerFi 
Finally, this is ready for your review.

YogevBokobza:
@thecode 
I think we are going to have a little issue in HA with this,
Right now shutter position and direction are single vars and here this is an array, Also for light in HA this is an Array (because of S11), and here this is single var.

Theoretically, this library is well-coded with those use cases, but in HA there will be a split in those use cases.
What do you think?

thecode:
> @thecode I think we are going to have a little issue in HA with this, Right now shutter position and direction are single vars and here this is an array, Also for light in HA this is an Array (because of S11), and here this is single var.
> 
> Theoretically, this library is well-coded with those use cases, but in HA there will be a split in those use cases. What do you think?

I don't understand the problem you are describing, these are well typed:
https://github.com/TomerFi/aioswitcher/blob/fca042f8a6f2eefff93561b530e80a31c220a1cb/src/aioswitcher/device/__init__.py#L348-L359

And in HA you check the type and add the correct number of entities:
https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/cover.py#L43-L47

https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/light.py#L41-L50

So you can just add another case similar to what we do in the switch platform:
https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/switch.py#L70-L73

Can you explain what is the problem we are going to face in HA?

YogevBokobza:
> > @thecode I think we are going to have a little issue in HA with this, Right now shutter position and direction are single vars and here this is an array, Also for light in HA this is an Array (because of S11), and here this is single var.
> > Theoretically, this library is well-coded with those use cases, but in HA there will be a split in those use cases. What do you think?
> 
> I don't understand the problem you are describing, these are well typed:
> 
> https://github.com/TomerFi/aioswitcher/blob/fca042f8a6f2eefff93561b530e80a31c220a1cb/src/aioswitcher/device/__init__.py#L348-L359
> 
> And in HA you check the type and add the correct number of entities: https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/cover.py#L43-L47
> 
> https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/light.py#L41-L50
> 
> So you can just add another case similar to what we do in the switch platform: https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/switch.py#L70-L73
> 
> Can you explain what is the problem we are going to face in HA?

This for example:
https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/cover.py#L89
it takes data.position but in this use case it should be something like data.position[Index]
You are right for the code you showed.

thecode:
> This for example: https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/cover.py#L89 it takes data.position but in this use case it should be something like data.position[Index] You are right for the code you showed.

I still don't understand what is the problem, you can subclass `SwitcherCoverEntity` in HA and add an index or you can check if the `data.position` is a list or not to know if you need to use an index.

YogevBokobza:
> > This for example: https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/switcher_kis/cover.py#L89 it takes data.position but in this use case it should be something like data.position[Index] You are right for the code you showed.
> 
> I still don't understand what is the problem, you can subclass `SwitcherCoverEntity` in HA and add an index or you can check if the `data.position` is a list or not to know if you need to use an index.

Yeah I guess you are right.. it's not a real problem but a matter of adding conditions to support both single value an array of values
It's not that nice but it can work

Anyway we go with this PR as it is and we can discuss this later in the PR for HA.
You can review this PR when you are ready.

src/aioswitcher/device/tools.py
```diff
@@ -263,6 +273,11 @@ def get_light_discovery_packet_index(
             return circuit_number
         else:
             raise ValueError("Invalid circuit number")
+    if device_type == DeviceType.RUNNER_S12:
+        if circuit_number == 0:
+            return circuit_number
+        else:
+            raise ValueError("Invalid circuit number")
```
thecode:
```suggestion
        if circuit_number != 0:
            raise ValueError("Invalid circuit number")
        return 0
```

src/aioswitcher/device/tools.py
```diff
@@ -263,6 +273,11 @@ def get_light_discovery_packet_index(
             return circuit_number
         else:
             raise ValueError("Invalid circuit number")
```
thecode: Please also  fix this part:
```python
if circuit_number not in [0,1]:
    raise ValueError("Invalid circuit number")
return circuit_number

File: src/aioswitcher/device/tools.py [hunk 3]
```diff
@@ -239,13 +243,19 @@ def get_shutter_discovery_packet_index(
     Used in retriving the shutter position/direction from the packet
     (based of device type and circuit number).
     """
-    if circuit_number != 0:
+    if (device_type != DeviceType.RUNNER_S12 and circuit_number != 0) or (
+        device_type == DeviceType.RUNNER_S12
+        and circuit_number != 0
+        and circuit_number != 1
+    ):
```
thecode: Try to improve readability for this guard (similar to my comments bellow)

src/aioswitcher/device/tools.py
YogevBokobza: It is done.

src/aioswitcher/device/tools.py
YogevBokobza: It is done.

src/aioswitcher/device/tools.py
YogevBokobza: It is done.

src/aioswitcher/device/tools.py
```diff
@@ -239,13 +243,19 @@ def get_shutter_discovery_packet_index(
     Used in retriving the shutter position/direction from the packet
     (based of device type and circuit number).
     """
-    if circuit_number != 0:
-        raise ValueError("Invalid circuit number")
+    if device_type != DeviceType.RUNNER_S12:
+        if circuit_number != 0:
+            raise ValueError("Invalid circuit number")
```
thecode:
```suggestion
    if device_type != DeviceType.RUNNER_S12 and circuit_number != 0:
            raise ValueError("Invalid circuit number")
```

src/aioswitcher/device/tools.py
```diff
@@ -239,13 +243,19 @@ def get_shutter_discovery_packet_index(
...
+    if device_type == DeviceType.RUNNER_S12:
+        if circuit_number not in [0, 1]:
+            raise ValueError("Invalid circuit number")
```
thecode: Merge into one if using `and`

src/aioswitcher/device/tools.py
```diff
@@ -259,10 +269,13 @@ def get_light_discovery_packet_index(
...
     if device_type == DeviceType.RUNNER_S11:
-        if circuit_number in [0, 1]:
-            return circuit_number
-        else:
+        if circuit_number not in [0, 1]:
             raise ValueError("Invalid circuit number")
+        return circuit_number
+    if device_type == DeviceType.RUNNER_S12:
+        if circuit_number != 0:
+            raise ValueError("Invalid circuit number")
```
thecode: Merge into one if using `and`

thecode (changes requested):

src/aioswitcher/device/tools.py
YogevBokobza: it is done

src/aioswitcher/device/tools.py
YogevBokobza:
``` 
    if device_type == DeviceType.RUNNER_S11:
        if circuit_number not in [0, 1]:
            raise ValueError("Invalid circuit number")
        return circuit_number
    if device_type == DeviceType.RUNNER_S12:
        if circuit_number != 0:
            raise ValueError("Invalid circuit number")
        return 0

    raise ValueError("only devices that has lights are allowed")
```

I cant make "and" here at it will break the logic.

src/aioswitcher/device/tools.py
YogevBokobza: it is done

thecode (approved):
Thanks!

YogevBokobza:
@TomerFi
Please approve, merge and make a release.
Thanks :)

TomerFi:
@YogevBokobza @thecode 
Sorry for the delay. If we set the release bump to auto, this will be a major release. Are we okay with that, or would you prefer me to manually set this to minor?

thecode:
I am ok with both options

YogevBokobza:
> @YogevBokobza @thecode Sorry for the delay. If we set the release bump to auto, this will be a major release. Are we okay with that, or would you prefer me to manually set this to minor?

Same. I am ok with both options.

TomerFi:
My bad - auto will be minor.

TomerFi:
Version [4.1.0](https://github.com/TomerFi/aioswitcher/releases/tag/4.1.0) released!

TomerFi:
I apologize for being a bit busy lately; thank you both for the hard work!

pshotts:
When will this be incorporated in the Home Assistant integration?

YogevBokobza:
> When will this be incorporated in the Home Assistant integration?

When this [pr](https://github.com/home-assistant/core/pull/127277) will be merged it will be part of HA.
Right now we are at block. Sadly, I am not sure if this will be any time soon..

pshotts:
Thanks for the update, and for all your work on this!

PR #801, opened 2024-08-17
fix: add aiohttp to deps

ayal:
## Description

add aiohttp to deps after this change: https://github.com/TomerFi/aioswitcher/commit/7c45f8bc3ed48e42df56f888a0ef09d4d6624a8a

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ exclude = [ ]
   [tool.poetry.dependencies]
   python = "^3.9.0"
   pycryptodome = ">=3.18.0"
-  aiohttp = ">3.10.3"
+  aiohttp = ">=3.10.3,<4.0.0"
 
 [tool.poetry.group.dev.dependencies]
 assertpy = "^1.1"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ exclude = [ ]
   [tool.poetry.dependencies]
   python = "^3.9.0"
   pycryptodome = ">=3.18.0"
-  aiohttp = "^3.10.3"
+  aiohttp = ">3.10.3"
 
 [tool.poetry.group.dev.dependencies]
 assertpy = "^1.1"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -23,6 +23,7 @@ exclude = [ ]
   [tool.poetry.dependencies]
   python = "^3.9.0"
   pycryptodome = ">=3.18.0"
+  aiohttp = "^3.10.3"
```
thecode:
```suggestion
  aiohttp = ">3.10.3"
```
Can we just make sure it is above the minimum version?

thecode (dismissed):

TomerFi:
@ayal, Before merging, can you please document what's wrong with version 4.0.0 of _aiohttp_ in the PR body?

thecode:
> @ayal, Before merging, can you please document what's wrong with version 4.0.0 of _aiohttp_ in the PR body?

I just noticed this was added in a later [commit](https://github.com/TomerFi/aioswitcher/pull/801/commits/0b4cdca4b7958a5a0868413524e6563e7a389204) after I approved the PR.
The main idea was to make sure we don't need to create a new release when `aiohttp` will reach version 4. 

As the one of the usages for `aiohttp` is going to be in Home Assistant and the most active developer in `aiohttp` is one of the Home Assistant core members I don't think there is a need to restrict anything. If there will be something breaking in `aiohttp` it will not be bumped in Home Assistant until all dependencies are compatible, if we restrict here we just force that we will not be compatible, but we don't know now anything that is not compatible.

ayal:
@thecode  @TomerFi
Changed it now to allow any version.
I'm used to a different version-pinning practice,
i.e pin to a specific version and then use some bot (for example renovate) for upgrading versions (by a pr to test the upgrade, then auto-merge if successful). Major versions are specially indicative to breaking changes so should be even more careful.

thecode (approved):

TomerFi (approved):

PR #804, opened 2024-10-06
feat: add Light SL01 and SL01 mini support

### LINKED ISSUE #696: Support for Swithcer smart light switches

Please Support Switcher smart light switches so that Home Assistant can control them
I opened an [issue ](https://github.com/home-assistant/core/issues/96790) over at HA , but was directed here

_No response_

_No response_

YogevBokobza:
## Description

> Add Switcher Light SL01 and SL01 Mini support

**Related issue (if any):** fixes #696 (Partially)

## Checklist

## Additional information

> Anything else?


src/aioswitcher/device/__init__.py
```diff
@@ -359,6 +374,17 @@ class SwitcherDualShutterSingleLightBase(ABC):
...
 
+@dataclass
+class SwitcherLightBase(ABC):
+    """Abstraction for all switcher devices controlling light.
+
+    Args:
+        lights: the current lights state.
+    """
+
+    lights: DeviceState
```
thecode:
```suggestion
    light: DeviceState
```
Maybe better call it `light` because it is not a list?

src/aioswitcher/device/__init__.py
```diff
@@ -359,6 +374,17 @@ class SwitcherDualShutterSingleLightBase(ABC):
     lights: DeviceState
```
thecode: Same here, missed in previous review, I think it might be better to call it `light`

src/aioswitcher/device/__init__.py
YogevBokobza: But this is already like this [here](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/device/__init__.py#L359).. Do you want to change this in the mentioned place as well?
The idea was to make things easier in HA integration.. less conditional stuff

Keep in mind that Light SL02 and SL03 will have 2 light and 3 lights modules

YogevBokobza: Maybe it's a good idea to have this var as `light` in all places? and it's value is single or array of DeviceState?

thecode: I agree it simply the usage of it later, but if we go this way I think we should align the shutter to also have a list for the `direction` and `position` even for single instance (`SwitcherSingleShutterDualLightBase`)

YogevBokobza: > I agree it simply the usage of it later, but if we go this way I think we should align the shutter to also have a list for the `direction` and `position` even for single instance (`SwitcherSingleShutterDualLightBase`)

What I mean is I am right now treating "lights", "position" and "direction" in HA as a single value or array of values, but if I set this variable to "light" then it will make things harder in HA since I have "lights" so I will need to have conditional for that as well.

I think the best logical way is to have "light", "position" and "direction" that can be a single value or array of values (Actually the best option was that all of those are arrays so it can be a single item in that or multiple, but this will be a headache to change now)

I think for this PR, we should go with "lights" because now it's aligned in HA and we make another PR that aligned all of the fields to be "light", "position" and "direction"

What do you think?

thecode: You already have an open PR to update HA for S12 so the naming is not a blocker, I would prefer we finish this PR, create another release and not hold back because it is already WIP in HA side.
About naming I am ok with `light` `position` and `direction` as lists as long as they are **all** single or plural and not mixed (of course they will all be lists even if it is a single item if it was not clear)

YogevBokobza: > You already have an open PR to update HA for S12 so the naming is not a blocker, I would prefer we finish this PR, create another release and not hold back because it is already WIP in HA side. About naming I am ok with `light` `position` and `direction` as lists as long as they are **all** single or plural and not mixed.

It is done.. changed all to `light`.
I didn't want to do this in this PR since it can become big PR and out of scope, but I go with your thought.

thecode: I think we have a broken discussion, looking at the changes they all have the same name but some hare now list and some a single item, in this case I don't agree with giving a light array and a single light the same name,.
My intention was that we align all classes that use `light`, `position` and `direction` to have a list and not a mix
```python
    position: List[int]
    direction: List[ShutterDirection]
    light: List[DeviceState]
```

Or:
```python
    positions: List[int]
    directions: List[ShutterDirection]
    lights: List[DeviceState]
```

YogevBokobza: > I think we have a broken discussion, looking at the changes they all have the same name but some hare now list and some a single item, in this case I don't agree with giving a light array and a single light the same name,. My intention was that we align all classes that use `light`, `position` and `direction` to have a list and not a mix
> 
> ```python
>     position: List[int]
>     direction: List[ShutterDirection]
>     light: List[DeviceState]
> ```
> 
> Or:
> 
> ```python
>     positions: List[int]
>     directions: List[ShutterDirection]
>     lights: List[DeviceState]
> ```

I can do that, but I must ask why the current implementation is so bad having a variable that can be a single value or array of values?
We figure out how to handle it in HA by separating the classes

thecode: > I can do that, but I must ask why the current implementation is so bad having a variable that can be a single value or array of values?
We figure out how to handle it in HA by separating the classes

So I will ask back, why do you want to give a variable that hold a single item and a variable that hold a list the same name? If you go this way name them `lights` if they have multiple and `light if it is a single item (same for the others).

My intention was to make it easier for you in HA

YogevBokobza: > > I can do that, but I must ask why the current implementation is so bad having a variable that can be a single value or array of values?
> > We figure out how to handle it in HA by separating the classes
> 
> So I will ask back, why do you want to give a variable that hold a single item and a variable that hold a list the same name? If you go this way name them `lights` if they have multiple and `light if it is a single item (same for the others).
> 
> My intention was to make it easier for you in HA

Because they have the same name, It is relatively easy to work with a variable called `light` that holds a single item or list of items but it is harder to work with 2 variables like `light` and `lights` this would make huge conditional to handle the cases.

So the second option you mentioned to have `light` and `lights` (same for position and direction) will make it harder.
If you don't agree with `light`, `position` and `direction` to holds a single item and list of items I will change this to the first option you mentioned before:
position: List[int]
direction: List[ShutterDirection]
light: List[DeviceState]

YogevBokobza: @thecode I can see you are busy and cannot answer to the above
I want to get progress here so I will start implement the option 1 you suggested:
```code
    position: List[int]
    direction: List[ShutterDirection]
    light: List[DeviceState]
```

YogevBokobza: It is done.

YogevBokobza:
@thecode I know you don't like to get pinged and you probably busy, but please try to give this PR attention so we can continue.
We have a little more devices to complete and the Switcher community is waiting :)
Thanks !

thecode (approved):

TomerFi:
@thecode @YogevBokobza 
Merged, release?

thecode:
> @thecode @YogevBokobza Merged, release?

Yes, thanks!

TomerFi:
Apologies for the delay; version [4.2.0](https://github.com/TomerFi/aioswitcher/releases/tag/4.2.0) was released.

TomerFi:
Wiki updated: https://github.com/TomerFi/aioswitcher/wiki

PR #808, opened 2024-10-31
chore: refactoring code

### LINKED ISSUE #760: Refactor src/aioswitcher/api/__init__.py

https://github.com/TomerFi/aioswitcher/pull/747#discussion_r1543027849

_No response_

_No response_

YogevBokobza:
## Description

**Related issue (if any):** fixes #760

## Checklist

## Additional information

> Anything else?


YogevBokobza:
@thecode @TomerFi 
Starting to refactor the base code to fix #760.
Regarding #760, I am unsure how to decrease this code more than I did here.

Any ideas would help.

src/aioswitcher/api/helper.py
```diff
@@ -0,0 +1,46 @@
...
+
+from ..device.tools import set_message_length, sign_packet_with_crc_key
+
+logger = getLogger(__name__)
+
+
+async def send_packet(
+    writer: StreamWriter, reader: StreamReader, packet_id: str, packet: str
+) -> bytes:
+    """Sign and send a packet, then read the response.
```
thecode: It would be better to add this as a private class method (`_send_packet`) under the `SwitcherApi`
- It is not used outside of this class
- As a member of the class it will have access to the `writer` & `reader` streams so you don't have to pass them

YogevBokobza: You are probably right, but it will miss the point when we want to decrease that file length.
If so, I must move this from helper to the same file Where SwitcherApi is.

thecode: With this PR you gather few methods which are used very often in a single method, this should reduce the size.
When we have all the common parts optimized in methods we can later try to split the transport into another class if it will make sense

YogevBokobza:
Not sure what else can be refactored. 
Maybe the exception for login in all the functions could move to the _login function? and then it will stop there if failed and in the other functions, we can continue from there, not checking if the login succeeded.

src/aioswitcher/api/__init__.py
```diff
@@ -153,6 +153,24 @@ async def disconnect(self) -> None:
             logger.info("switcher device not connected")
         self._connected = False
 
+    async def _send_packet(self, packet_id: str, packet: str) -> bytes:
```
TomerFi, line 156: If we only use the _packet_id_ to log a debug message, let's log it before invoking __send_packet_.

TomerFi (changes requested):

TomerFi:
> Not sure what else can be refactored.
@YogevBokobza How about:
1. Move [COMMAND_TO_MODE](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/api/remotes.py#L30) after [MODE_TO_COMMAND](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/api/remotes.py#L38) and change it to:

TomerFi:
> Not sure what else can be refactored.

@YogevBokobza, a couple of ideas:

In [api/remotes.py](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/api/remotes.py)
1. move _COMMAND_TO_MODE_ after _MODE_TO_COMMAND_, and change it to:
```python
COMMAND_TO_MODE = {v: k for k, v in MODE_TO_COMMAND.items()}
```
2. move _COMMAND_TO_FAN_LEVEL_ after _FAN_LEVEL_TO_COMMAND_, and change it to:
```python
COMMAND_TO_FAN_LEVEL = {v: k for k, v in FAN_LEVEL_TO_COMMAND.items()}
```

---

In [device/tools.py](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/device/tools.py#L149) about the _convert_str_to_devicetype_ function:
1. First of all, it's only used by the control device script. It shouldn't be part of the package. Let's move it to the  [scripts/control_device.py](https://github.com/TomerFi/aioswitcher/blob/dev/scripts/control_device.py#L778).

2. This entire function is redundant. It could be a simple dict, much like we did for the [Switcher WebAPI project](https://github.com/TomerFi/switcher_webapi/blob/dev/app/webapp.py#L70):
```python
DEVICES = {
    DeviceType.MINI.value: DeviceType.MINI,
    DeviceType.POWER_PLUG.value: DeviceType.POWER_PLUG,
    ...
}
```
> [!NOTE]
> Side effects of 1+2 are that we won't need the test cases and get rid of the default "Mini" device, which sounds wrong.

3. I suggest changing the script and the abovementioned dict. Currently, users need to pass the exact name of the device, for instance: _"Switcher Light SL02 Mini"_. That's too harsh! I would use some translation, something like, again, we did for the [Switcher WebAPI project](https://github.com/TomerFi/switcher_webapi/blob/dev/app/webapp.py#L70). If not, I would lowercase both ends to allow some flexibility. But this could be in a different PR.

TomerFi:
> 3\. I suggest changing the script and the abovementioned dict. Currently, users need to pass the exact name of the device, for instance: _"Switcher Light SL02 Mini"_. That's too harsh! I would use some translation, something like, again, we did for the [Switcher WebAPI project](https://github.com/TomerFi/switcher_webapi/blob/dev/app/webapp.py#L70). If not, I would lowercase both ends to allow some flexibility. But this could be in a different PR.

To clarify, by lowercase both ends, I mean:
1. Change the dict to
```python
DEVICES = {
    DeviceType.MINI.value.lower(): DeviceType.MINI,
    DeviceType.POWER_PLUG.value.lower(): DeviceType.POWER_PLUG,
    ...
}
```
2. And change the script to:
```python
args.device_type = DEVICES(args["device_type"].lower())
```

The part I wrote about the translation can be done in a different PR.

TomerFi:
Also, if we're refactoring stuff in the _control_device_ script. I would change this statement, too:
```python
args.device_type = ...
```
Something about rewriting the values parsed from the command line arguments feels wrong. I would save this as a variable instead.

src/aioswitcher/api/__init__.py
YogevBokobza: The idea was to decrease file size.. If I do that it will increase this line again and again.

YogevBokobza: @TomerFi  what do you think?

YogevBokobza:
> > Not sure what else can be refactored.
> 
> @YogevBokobza, a couple of ideas:
> 
> In [api/remotes.py](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/api/remotes.py)
> 
> 1. move _COMMAND_TO_MODE_ after _MODE_TO_COMMAND_, and change it to:
> 
> ```python
> COMMAND_TO_MODE = {v: k for k, v in MODE_TO_COMMAND.items()}
> ```
> 
> 2. move _COMMAND_TO_FAN_LEVEL_ after _FAN_LEVEL_TO_COMMAND_, and change it to:
> 
> ```python
> COMMAND_TO_FAN_LEVEL = {v: k for k, v in FAN_LEVEL_TO_COMMAND.items()}
> ```
> 
> In [device/tools.py](https://github.com/TomerFi/aioswitcher/blob/dev/src/aioswitcher/device/tools.py#L149) about the _convert_str_to_devicetype_ function:
> 
> 1. First of all, it's only used by the control device script. It shouldn't be part of the package. Let's move it to the  [scripts/control_device.py](https://github.com/TomerFi/aioswitcher/blob/dev/scripts/control_device.py#L778).
> 2. This entire function is redundant. It could be a simple dict, much like we did for the [Switcher WebAPI project](https://github.com/TomerFi/switcher_webapi/blob/dev/app/webapp.py#L70):
> 
> ```python
> DEVICES = {
>     DeviceType.MINI.value: DeviceType.MINI,
>     DeviceType.POWER_PLUG.value: DeviceType.POWER_PLUG,
>     ...
> }
> ```
> 
> Note
> 
> Side effects of 1+2 are that we won't need the test cases and get rid of the default "Mini" device, which sounds wrong.
> 
> 3. I suggest changing the script and the abovementioned dict. Currently, users need to pass the exact name of the device, for instance: _"Switcher Light SL02 Mini"_. That's too harsh! I would use some translation, something like, again, we did for the [Switcher WebAPI project](https://github.com/TomerFi/switcher_webapi/blob/dev/app/webapp.py#L70). If not, I would lowercase both ends to allow some flexibility. But this could be in a different PR.

It is done. translation can be done in this PR or the next.
The question is if this makes this PR bigger than it is..

YogevBokobza:
> > 3. I suggest changing the script and the abovementioned dict. Currently, users need to pass the exact name of the device, for instance: _"Switcher Light SL02 Mini"_. That's too harsh! I would use some translation, something like, again, we did for the [Switcher WebAPI project](https://github.com/TomerFi/switcher_webapi/blob/dev/app/webapp.py#L70). If not, I would lowercase both ends to allow some flexibility. But this could be in a different PR.
> 
> To clarify, by lowercase both ends, I mean:
> 
> 1. Change the dict to
> 
> ```python
> DEVICES = {
>     DeviceType.MINI.value.lower(): DeviceType.MINI,
>     DeviceType.POWER_PLUG.value.lower(): DeviceType.POWER_PLUG,
>     ...
> }
> ```
> 
> 2. And change the script to:
> 
> ```python
> args.device_type = DEVICES(args["device_type"].lower())
> ```
> 
> The part I wrote about the translation can be done in a different PR.

It is done.

YogevBokobza:
> Also, if we're refactoring stuff in the _control_device_ script. I would change this statement, too:
> 
> ```python
> args.device_type = ...
> ```
> 
> Something about rewriting the values parsed from the command line arguments feels wrong. I would save this as a variable instead.

It is done.

YogevBokobza:
@thecode @TomerFi Anything else you think need refactoring?

src/aioswitcher/api/__init__.py
TomerFi: I still think it should be outside the function. But it's not a deal breaker for me. We can go this way too.

TomerFi (approved):

thecode (approved):

PR #821, opened 2024-12-28
fix: cleared multiple deperecation notices

TomerFi:
## Description

When I ran some tests, I noticed a couple of deprecation notices:
- For `datetime.datetime.utcnow()`, the resolution is specified in the deprecation notice: `datetime.datetime.now(datetime.UTC)`.
- For `ssl.SSLContext()`, since version 3.10, [SSLContext](https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext), the default value for the SSLContext function, [PROTOCOL_TLS](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS), was deprecated in favour of  [PROTOCOL_TLS_CLIENT](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT) and [PROTOCOL_TLS_SERVER](https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS_SERVER).

image

## Checklist

## Additional information

image

### DIFF
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -174,7 +174,7 @@ async def validate_token(username: str, token: str) -> bool:
     request_data = {"email": username, "token": token}
     is_token_valid = False
     # Preload the SSL context
-    ssl_context = ssl.SSLContext()
+    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
 
     logger.debug("calling API call for Switcher to validate the token")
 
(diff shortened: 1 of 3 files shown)


thecode:
@YogevBokobza would be nice if you can validate the changed method in this PR using a real device

TomerFi:
> @YogevBokobza would be nice if you can validate the changed method in this PR using a real device

@thecode, I added related information to the PR's body. I should have done that when I opened the PR.

thecode (approved):

PR #833, opened 2025-01-16
feat: time based power plug

### LINKED ISSUE #335: Switcher power plug with firmware >= 1.46 supports timed base capabilities
### What did you had in mind?

According to research performed by @thecode and @dmatik,
`Switcher power plugs` with firmware >= 1.46 supports timed based capabilities such as:

- Auto shutdown of the device based on the configured value.
- Displaying the remaining time for the current run.

These are capabilities that up until now were used in `Switcher water heater devices`.
It should be stated, power plugs with firmware < 1.46, return a default value for the remaining time,
So any modification made for this enhancement, should not affect backward compatibility.

Power plugs with the firmware of >= 1.46 come with the configuration option for auto-shutdown.
This configuration is not visible to this integration.

We can split this into two workflows:

1. Setting the device configuration auto-shutdown value ~~and/or turning it on with a designated timer~~ (timer is not supported see [comment](https://github.com/TomerFi/aioswitcher/issues/335#issuecomment-1007633888),) should probably work out of the box, as the API context doesn't verify the type of the device before sending out the required packet. So I don't think any modifications are required for this part.

2. The device state should display the new capabilities value.
    This will require some modifications, so let's dive in:
  - The [DeviceCategory Enum](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L25) is used to categorized the device type.
  - The [DeviceType Enum](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L33) helps us identify the discovered device.
  - The [DeviceState Enum](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L70) although, not quite related to this issue,  but worth mentioning, helps us categorized the device's state.

-------

  -  The [SwitcherBase class](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L95) is the abstraction all switcher devices inherit from.
  - The [SwitcherPowerBase class](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L122) is an abstraction used for switcher devices with the capability to have `power_consumption` and `electric_current` properties.
  - The [SwitcherTimedBase class](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L136) is an abstraction used for switcher devices with the capability to have `remaining_time` and `auto_shutdown` properties.

-------

  - The [SwitcherPowerPlug class](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L156) is a concrete implementation of the switcher power plug, it inherits from both `SwitcherPowerBase` and `SwitcherBase`.
  - The [SwitcherWaterHeater class](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/device/__init__.py#L172) is a concrete implementation of the switcher water heater, it inherits from `SwitcherTimedBased`, `SwitcherPoerBase`, and `SwitcherBase`.

-------

  - The [_parse_device_from_datagram](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/bridge.py#L42) private method is the one place that identifies the different device types.
  - The [SwitcherBridge](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/bridge.py#L107) is the udp context manager that utilizes the above `_parse_device_from_datagram` in order to return the correct device.

-------

Although, as for this enhancement, the power plugs and water heater are basically the same devices.
**I wouldn't recommend dropping** the separation between them,
we can just add capabilities to the power plug representation and make it similar to the water heater one.

-------

In order to give the power plug timed-based capabilities, we simply need to make the `SwitcherPowerPlug` class inherit from `SwitcherTimedBased` as well as the current inheritance from `SwitcherPowerBase` and `SwitcherBase`.
Just remember - **it's python - order matters** so keep it in line with the `SwitcherWaterHeater` class.

We then need to add the two new properties when discovering a power plug device,
in `_parse_device_from_datagram`,
find the [power plug if branch](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/bridge.py#L88) and add the new properties,
it should be quite similar to the [water heater if branch](https://github.com/TomerFi/aioswitcher/blob/5715b7addf5aa044cec5a33c0e6890f084722187/src/aioswitcher/bridge.py#L67).

These modifications are allegedly not that big or complicated,
But I assume testing modification will be required as well.

YogevBokobza:
## Checklist

### DIFF
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -233,7 +233,7 @@ class SwitcherStateResponse(SwitcherBaseResponse):
         parser = StateMessageParser(self.unparsed_response)
 
         self.state = parser.get_state()
-        self.time_left = parser.get_time_left()
+        self.time_left = parser.get_time_left() if self.state == DeviceState.ON and parser.get_auto_shutdown() != "00:00:00" else "00:00:00"
         self.time_on = parser.get_time_on()
         self.auto_shutdown = parser.get_auto_shutdown()
         self.power_consumption = parser.get_power_consumption()
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -233,7 +233,11 @@ class SwitcherStateResponse(SwitcherBaseResponse):
         parser = StateMessageParser(self.unparsed_response)
 
         self.state = parser.get_state()
-        self.time_left = parser.get_time_left() if self.state == DeviceState.ON and parser.get_auto_shutdown() != "00:00:00" else "00:00:00"
+        self.time_left = (
+            parser.get_time_left()
+            if self.state == DeviceState.ON and parser.get_auto_shutdown() != "00:00:00"
+            else "00:00:00"
+        )
         self.time_on = parser.get_time_on()
         self.auto_shutdown = parser.get_auto_shutdown()
         self.power_consumption = parser.get_power_consumption()
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -137,7 +137,8 @@ def _parse_device_from_datagram(
                     electric_current,
                     (
                         parser.get_remaining()
-                        if device_state == DeviceState.ON and parser.get_auto_shutdown() != "00:00:00"
+                        if device_state == DeviceState.ON
+                        and parser.get_auto_shutdown() != "00:00:00"
                         else "00:00:00"
                     ),
                     parser.get_auto_shutdown(),
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -135,6 +135,12 @@ def _parse_device_from_datagram(
                     device_type.token_needed,
                     power_consumption,
                     electric_current,
+                    (
+                        parser.get_remaining()
+                        if device_state == DeviceState.ON
+                        else "00:00:00"
+                    ),
+                    parser.get_auto_shutdown(),
                 )
             )
 
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -137,7 +137,7 @@ def _parse_device_from_datagram(
                     electric_current,
                     (
                         parser.get_remaining()
-                        if device_state == DeviceState.ON
+                        if device_state == DeviceState.ON and parser.get_auto_shutdown() != "00:00:00"
                         else "00:00:00"
                     ),
                     parser.get_auto_shutdown(),
--- a/src/aioswitcher/device/__init__.py
+++ b/src/aioswitcher/device/__init__.py
@@ -436,7 +436,7 @@ class SwitcherLightBase(ABC):
 
 @final
 @dataclass
-class SwitcherPowerPlug(SwitcherPowerBase, SwitcherBase):
+class SwitcherPowerPlug(SwitcherTimedBase, SwitcherPowerBase, SwitcherBase):
     """Implementation of the Switcher Power Plug device.
 
     Please Note the order of the inherited classes to understand the order of the
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -44,10 +44,6 @@ def seconds_to_iso_time(all_seconds: int) -> str:
     """
     minutes, seconds = divmod(int(all_seconds), 60)
     hours, minutes = divmod(minutes, 60)
-    if hours > 23:
-        hours = 0
-        minutes = 0
-        seconds = 0
 
     return datetime.time(hour=hours, minute=minutes, second=seconds).isoformat()
 
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -44,6 +44,10 @@ def seconds_to_iso_time(all_seconds: int) -> str:
     """
     minutes, seconds = divmod(int(all_seconds), 60)
     hours, minutes = divmod(minutes, 60)
+    if hours > 23:
+        hours = 0
+        minutes = 0
+        seconds = 0
 
     return datetime.time(hour=hours, minute=minutes, second=seconds).isoformat()
 
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -44,6 +44,8 @@ def seconds_to_iso_time(all_seconds: int) -> str:
     """
     minutes, seconds = divmod(int(all_seconds), 60)
     hours, minutes = divmod(minutes, 60)
+    if hours > 24:
+        hours = 0
 
     return datetime.time(hour=hours, minute=minutes, second=seconds).isoformat()
 
--- a/tests/test_api_messages.py
+++ b/tests/test_api_messages.py
@@ -66,9 +66,10 @@ def test_switcher_state_response_dataclass(get_power_consumption, get_auto_shutd
     assert_that(sut.power_consumption).is_equal_to(1640)
     assert_that(sut.electric_current).is_equal_to(7.5)
 
-    for mock_method in [get_power_consumption, get_auto_shutdown, get_time_on, get_time_left, get_state]:
+    for mock_method in [get_power_consumption, get_time_on, get_time_left, get_state]:
         mock_method.assert_called_once()
-
+    get_auto_shutdown.assert_called()
+    assert get_auto_shutdown.call_count == 2
 
 @patch(messages.__name__ + ".get_schedules", return_value={Mock(), Mock()})
 def test_switcher_get_schedules_response_dataclass_with_two_schedules(get_schedules):
--- a/tests/test_api_messages.py
+++ b/tests/test_api_messages.py
@@ -71,6 +71,7 @@ def test_switcher_state_response_dataclass(get_power_consumption, get_auto_shutd
     get_auto_shutdown.assert_called()
     assert get_auto_shutdown.call_count == 2
 
+
 @patch(messages.__name__ + ".get_schedules", return_value={Mock(), Mock()})
 def test_switcher_get_schedules_response_dataclass_with_two_schedules(get_schedules):
     sut = SwitcherGetSchedulesResponse(b'moot binary data2')
--- a/tests/test_device_dataclasses.py
+++ b/tests/test_device_dataclasses.py
@@ -115,6 +115,8 @@ def test_given_a_device_of_type_power_plug_when_instantiating_as_a_power_plug_sh
         fake_data.token_needed,
         fake_data.power_consumption,
         fake_data.electric_current,
+        fake_data.remaining_time,
+        fake_data.auto_shutdown,
     )
 
     assert_that(sut.device_type).is_equal_to(DeviceType.POWER_PLUG)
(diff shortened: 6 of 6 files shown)


TomerFi:
I'm removing the Stale action because it's not working as expected.

TomerFi:
Any update about this one @YogevBokobza @thecode ?

thecode:
> Any update about this one @YogevBokobza @thecode ?

My status is the same as reported at https://github.com/TomerFi/aioswitcher/issues/335#issuecomment-3009197126

I don't know how to progress with this PR since the workaround trying to fix the problem with devices that report an 30:00:00 when the timer is disabled did not work on my device last time I checked. 

I think @YogevBokobza need to get information from Switcher about this or alternatively a plug that use this value and fix the PR.

TomerFi:
@thecode I just want to make sure I understand, it's been a while since we discussed this. This PR works for @YogevBokobza 's device, but not for yours? Do we know if there's any difference between the devices? Maybe different firmwares?

thecode:
> @thecode I just want to make sure I understand, it's been a while since we discussed this. This PR works for @YogevBokobza 's device, but not for yours? Do we know if there's any difference between the devices? Maybe different firmwares?

See https://github.com/TomerFi/aioswitcher/issues/335#issuecomment-2666816201 there is a difference in the firmware version, but since these devices can't be upgraded anymore and we don't know how to read the firmware version from them we must have a solution that will at least not raise an exception when an older device exists on the network.

TomerFi:
> but since these devices can't be upgraded anymore and we don't know how to read the firmware version

@thecode So old devices can't upgrade their firmware? Wo we know what's behind this, is it like an old device not supported by the vendor anymore or something?

> we must have a solution that will at least not raise an exception when an older device exists on the network.

@YogevBokobza Can we do that? Identify the device in question is an old device and log something?

thecode:
> @thecode So old devices can't upgrade their firmware? Wo we know what's behind this, is it like an old device not supported by the vendor anymore or something?

Switcher policy is that devices in warranty are automatically updated and out of warranty are never updated (at least this is what they told us).

When we considered adding this feature we already had a problem, there are already devices with firmware bellow 1.46 which doesn't support time capabilities, now we have 3 types:
1. <1.46 doesn't support time capabilities
2. ==1.46 support with certain API
3. \>1.46 support with different API

In that time Switcher promised me they will updated all devices to 1.46 so we don't have to deal with different firmware support, this didn't happen, adding to that we didn't know at that time about another API change which will happen in the future.

We don't have a way to know the firmware version from the device (at least what they told me) and users can't update their devices, so to be honest I think we should drop this feature.

@YogevBokobza if Switcher can supply you the data of how to know the device firmware version (locally not from a cloud server) we can add a flag for each device if it supports it or not (in the table bellow), otherwise I suggest to drop this feature.

https://github.com/TomerFi/aioswitcher/blob/f087f18794259cfc33751447316f625ae91714d0/src/aioswitcher/device/__init__.py#L38-L49

YogevBokobza:
> > @thecode So old devices can't upgrade their firmware? Wo we know what's behind this, is it like an old device not supported by the vendor anymore or something?
> 
> Switcher policy is that devices in warranty are automatically updated and out of warranty are never updated (at least this is what they told us).
> 
> When we considered adding this feature we already had a problem, there are already devices with firmware bellow 1.46 which doesn't support time capabilities, now we have 3 types:
> 
> 1. <1.46 doesn't support time capabilities
> 2. ==1.46 support with certain API
> 3. >1.46 support with different API
> 
> In that time Switcher promised me they will updated all devices to 1.46 so we don't have to deal with different firmware support, this didn't happen, adding to that we didn't know at that time about another API change which will happen in the future.
> 
> We don't have a way to know the firmware version from the device (at least what they told me) and users can't update their devices, so to be honest I think we should drop this feature.
> 
> @YogevBokobza if Switcher can supply you the data of how to know the device firmware version (locally not from a cloud server) we can add a flag for each device if it supports it or not (in the table bellow), otherwise I suggest to drop this feature.
> 
> https://github.com/TomerFi/aioswitcher/blob/f087f18794259cfc33751447316f625ae91714d0/src/aioswitcher/device/__init__.py#L38-L49

I been told same as you are that it's impossible to get firmware version from  the device itself, its not hardcoded into the packets the devices send or receive..
Even if they do implement this older devices won't have it..
So I am not sure what to do here.. :|

Maybe an API to get the firmware version is possible.. but I am not sure if that's an option.

TomerFi:
@thecode Thanks for the detailed explanation. I understand the situation now.
Given that Switcher doesn't support out-of-warranty devices and there's no way to detect firmware version locally, I think we need to decide on the project's direction.

My view is we should support current firmware and move forward with new features. Here's why:

- The vendor doesn't support these older devices - they won't update them or provide detection methods
- New users with current firmware expect features to work
- Without moving forward, we're perpetually limited by unsupported hardware
- I'd rather see the project support current capabilities than drop features because of constraints from unsupported devices. 

What do you think?

thecode:
> @thecode Thanks for the detailed explanation. I understand the situation now. Given that Switcher doesn't support out-of-warranty devices and there's no way to detect firmware version locally, I think we need to decide on the project's direction.
> 
> My view is we should support current firmware and move forward with new features. Here's why:
> 
> * The vendor doesn't support these older devices - they won't update them or provide detection methods
> * New users with current firmware expect features to work
> * Without moving forward, we're perpetually limited by unsupported hardware
> * I'd rather see the project support current capabilities than drop features because of constraints from unsupported devices.
> 
> What do you think?

I disagree, we can't break older devices, my last test with this device created an exception on runtime which totally break discovery (meaning even any other device will not work), however even if we assume this is fixed, if we don't have a way to detect the firmware I don't see any way we can move forward because even for that we need a way to tell the user that the device is not supported or does not support certain feature.

What would happen with older devices when something doesn't work and the user doesn't know why? eventually that falls on me. I asked @YogevBokobza to add himself as a code owner for Switcher at Home Assistant so he can help with issues (Home Assistant bot tag code owners for help), but eventually since he added himself (2.5 years ago) I am the only one trying (mostly without success due to lake of feedback from the logs and I don't have devices) help with issues.

This feature is still controlled by the Switcher app for those who want it and they can still operate the device from Home Assistant. We also don't have any data when Switcher updated to which firmware so the I don't know how to address the term "older devices", people may still be using these devices and they need to work. 

Regardless of that, I mostly see requests to support new devices much more than this specific feature for this device, if you look at the linked issue https://github.com/TomerFi/aioswitcher/issues/335 it is only us discussing there, not a single user request of when does this implemented (same result in Facebook/forums) but if you look at support for new devices there are users asking about them.

We introduced in the past a breaking change for Shelly, we stopped support for device with very old firmware (more than 7 years which is not the case here) this alone made many users unhappy although we clearly indicated that the device use non-supported firmware and even allowed the user to update to latest firmware from within Home Assistant directly.

TomerFi:
@thecode I understand your concern about backward compatibility, and I recognize you're speaking from your experience maintaining both the Switcher and Shelly integrations in Home Assistant.

However, I think there's an important distinction between the two. Shelly is distributed globally with a massive international user base - breaking changes there create significant support burden and complaints travel fast through that large community. Switcher, on the other hand, is Israel-only with a much smaller, localized user base. This difference matters.

For Switcher's smaller community:
- Breaking changes are more manageable
- Communication about changes propagates more easily
- Users can more directly address upgrade issues with the vendor in a single market

I don't think we should limit ourselves to only supporting features that work with old firmware. By doing so, we're taking on a support burden that the vendor explicitly doesn't maintain. This might be one feature now, but it can grow exponentially. Eventually, we risk ending up with an outdated integration that primarily supports legacy devices while falling behind on current capabilities.

Users frustrated by this can direct that feedback to the vendor - maybe it will prompt them to change their approach or provide upgrade paths.

That said, if we currently have a minimal number of users requesting this feature and they have a workaround via the vendor's app, I'm open to closing this PR without merging (sorry @YogevBokobza ). We could open an issue to track interest, and if we see more users requesting it, we can reevaluate.

What do you think about that approach?

TomerFi:
@thecode One more thing - I realize that as the Home Assistant integration grows, the support burden on you increases. 

I stepped away from maintaining the Home Assistant integration when I moved out of Israel and stopped using Home Assistant. While I still can't physically test with Switcher devices (not available where I live), I can set up Home Assistant if needed.

If the support load is becoming overwhelming, I'd be happy to be re-added as a co-maintainer to help where I can - handling issues, reviewing code, managing releases, etc. You'd remain the primary maintainer with hands-on device access, but I could help lighten the load on the integration side.

Just wanted to put that out there in case it's useful.

YogevBokobza:
@TomerFi @thecode 
Following your comments here, I switch this PR to Draft.
Feel free to close it.

thecode:
> If the support load is becoming overwhelming, I'd be happy to be re-added as a co-maintainer to help where I can - handling issues, reviewing code, managing releases, etc. You'd remain the primary maintainer with hands-on device access, but I could help lighten the load on the integration side.

You are very welcome to add yourself as a code owner any time but note that:
1. When there are multiple code owners reviewers usually wait for others to give feedback, so if someone is added without being active it may slow down (which is why I originally asked to remove you)
2. Simple issues are not a problem, I manage them, the issues I can't help are things that need to close a loop with Switcher or test the behavior on a real device.

These are currently two issues:
- https://github.com/home-assistant/core/issues/157523 is waiting for fix/feedback from Switcher, which I assume if it didn't happen for few years it won't be fixed. We can try to bypass it by using the device IP from the UDP header and not from the payload as we do now.

https://github.com/TomerFi/aioswitcher/blob/dab53506fc1d3a7255329e39eec20009ec7dece0/src/aioswitcher/bridge.py#L511-L521

- https://github.com/home-assistant/core/issues/155523 needs Switcher attention to understand if it is a problem with the device or firmware compatibility problem.

TomerFi:
@thecode 

> When there are multiple code owners reviewers usually wait for others to give feedback, so if someone is added without being active it may slow down (which is why I originally asked to remove you)

Regarding the code ownership, unless I am misremembering, I don’t believe we ever discussed my removal prior to me initiating [PR #92277](https://github.com/home-assistant/core/pull/92277). I chose to step away because I had stopped using Home Assistant a couple of years prior to moving, and once I relocated, I no longer had access to the hardware for testing.

As for the concern that multiple reviewers slow down progress: while that might be the case now, it definitely wasn’t true during the three years I was an inactive maintainer. During that time, you made huge progress and rewrote a lot of the code without ever being halted by my lack of responsiveness. This shows that having me as a code owner was never a bottleneck, so it couldn't have been a reason I was 'asked to leave.'

While those circumstances haven't changed, my offer to reinstall the software was a genuine attempt to help with the software-side support burden. However, based on your response, it seems my help is neither needed nor appreciated. Since the issues you linked are hardware-dependent, I agree that I wouldn't be able to assist with those specifically. I'm fine with that and will keep my time focused on other projects.

Regarding this PR: while I appreciate everything you do for the Home Assistant community, I hope we can keep personal preferences separate from the technical direction of the library. We cannot indefinitely block new features to support devices that even the vendor has phased out.

I’ve agreed to hold off for now since there is a viable workaround and limited user demand, but I’m asking you—as a maintainer closer to the user base—to raise a flag when this becomes a broader requirement. I trust you to do the right thing for the project when that time comes. Ultimately, blocking changes here because of how the Shelly community reacted to unrelated breaking changes isn't a viable long-term argument. The Shelly and Switcher user bases are completely different—Switcher is effectively an Israel-only product, whereas Shelly is a global brand with a massive audience. You cannot compare the magnitude or the expectations of the two.

PR #857, opened 2026-03-05
fix: remove custom SSL context from validate_token

### LINKED ISSUE #856: SSL verification fails in `validate_token()` due to empty `SSLContext` (missing CA certificates)
### What happened?

Hi there,

Thanks for this amazing python lib, getting to manage switcher devices programmatically is just a blessing 🙏 

When configuring the Switcher integration in Home Assistant, token
validation fails with an SSL verification error:

    ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate

This happens during the `validate_token()` call in
`aioswitcher.device.tools`.

------------------------------------------------------------------------

## Environment

- Home Assistant Core: 2026.3.0 (HA OS)
- Python: 3.14 (inside the Home Assistant container)
- `aioswitcher`: version bundled with Home Assistant
- Installation type: Home Assistant OS (containerized Core)

------------------------------------------------------------------------

## Error trace

Relevant section of the stack trace:

    File "/usr/local/lib/python3.14/site-packages/aioswitcher/device/tools.py", line 182, in validate_token
      async with session.post(
          request_url, data=request_data, ssl=ssl_context
      ) as response:

    aiohttp.client_exceptions.ClientConnectorCertificateError:
    Cannot connect to host switcher.co.il:443 ssl:True
    [SSLCertVerificationError: certificate verify failed: unable to get local issuer certificate]

------------------------------------------------------------------------

## Root cause

`validate_token()` constructs a custom SSL context:

    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)

This creates a **blank SSL context without loading any CA
certificates**, so TLS verification fails when `aiohttp` tries to
validate the server certificate.

In contrast, libraries like `requests` work correctly because they load
trusted CA certificates (for example via `certifi`).

Because `aiohttp` is explicitly passed `ssl=ssl_context`, it bypasses
Python's default trust configuration and uses the empty context,
resulting in the verification error.

------------------------------------------------------------------------

## Workaround

Replacing the SSL context creation with a default context that loads
system CA certificates resolves the issue.

Manual patch inside the Home Assistant container:

    sed -i 's/ssl_context = ssl\.SSLContext(ssl\.PROTOCOL_TLS_CLIENT)/ssl_context = ssl.create_default_context()/' /usr/local/lib/python3.14/site-packages/aioswitcher/device/tools.py

After restarting Home Assistant, token validation works correctly.

------------------------------------------------------------------------

## Proposed fix

Replace the current SSL context initialization:

    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)

with:

    ssl_context = ssl.create_default_context()

This ensures the context loads the system's trusted CA certificates
while preserving certificate verification.

Alternatively, the SSL parameter could be removed entirely and aiohttp's
default SSL handling could be used.

------------------------------------------------------------------------

## Why this fix is correct

`ssl.create_default_context()`: - Enables certificate verification -
Loads trusted root CAs - Uses Python's recommended TLS configuration

The current implementation unintentionally creates a context that cannot
validate any certificate chain.

------------------------------------------------------------------------

## Additional notes

This issue is reproducible in containerized environments such as **Home
Assistant OS**, where Python otherwise correctly validates certificates
using the system or `certifi` CA bundle.

Switching to `ssl.create_default_context()` resolves the issue without
affecting functionality.

### Module Version

version bundled with Home Assistant

### Device Type

Switcher V2

### Firmware Version

|

### Relevant log output

```shell

```

TomerFi:
## Summary

- Remove the custom `SSLContext` from `validate_token()` and let aiohttp use its default SSL handling.
- The custom context was created with `PROTOCOL_TLS_CLIENT` (enabling certificate verification) but without loading any CA certificates, causing all TLS handshakes to fail with `CERTIFICATE_VERIFY_FAILED`.
- The `ssl` import is also removed as it is no longer used.

## Why this works

aiohttp's `ssl` parameter on request methods [defaults to `True`](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession.request), which is documented as:

> `ssl` – SSL validation mode. `True` for default SSL check (`ssl.create_default_context()` is used), `False` for skip SSL certificate validation, `aiohttp.Fingerprint` for fingerprint validation, `ssl.SSLContext` for custom SSL certificate validation.

By not passing `ssl=` at all (defaulting to `True`), aiohttp internally calls `ssl.create_default_context()`, which loads system CA certificates and enables proper verification — exactly the correct behavior.

Fixes #856

### DIFF
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -15,7 +15,6 @@
 """Switcher integration device module tools."""
 
 import datetime
-import ssl
 import time
 from base64 import b64decode
 from binascii import crc_hqx, hexlify, unhexlify
@@ -173,15 +172,11 @@ async def validate_token(username: str, token: str) -> bool:
     request_url = "https://switcher.co.il/ValidateToken/"
     request_data = {"email": username, "token": token}
     is_token_valid = False
-    # Preload the SSL context
-    ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
 
     logger.debug("calling API call for Switcher to validate the token")
 
     async with aiohttp.ClientSession() as session:
-        async with session.post(
-            request_url, data=request_data, ssl=ssl_context
-        ) as response:
+        async with session.post(request_url, data=request_data) as response:
             if response.status == 200:
                 logger.debug("request successful")
                 try:


TomerFi:
FYI @thecode @YogevBokobza — this bug was introduced in aioswitcher 6.0.1 (Jan 2025), but HA stayed on 6.0.0 until the heater support PR (home-assistant/core#162588) bumped it to 6.1.0 on Feb 13. That 13-month gap is why reports only started showing up now with HA 2026.3.

The impact is limited to **new setups or re-configurations** of token-based devices (Runner S11/S12, Lights, Heater) — `validate_token()` is only called during the config flow, so existing integrations configured before HA 2026.3 are unaffected.

Once this is merged and released, a version bump PR in HA core will be needed to pick up the fix.

YogevBokobza (approved):

thecode (approved):
I tested and token validates successfully with this change

PR #887, opened 2026-07-01
fix: set SO_REUSEADDR on the broadcast listener socket

nitaybz:
## Problem

The UDP bridge can fail to restart with `OSError: [Errno 98] Address already in use`. Downstream in the Home Assistant `switcher_kis` integration this shows up as devices going unavailable and the integration wedging on reload until a full Home Assistant restart: when the entry reloads, the previous listener socket is still lingering in `TIME_WAIT`, so rebinding the same broadcast port fails.

## The fix

Create each broadcast listener socket manually with `SO_REUSEADDR` set before `bind`, then hand it to `create_datagram_endpoint` via `sock=`, so a reload can rebind the port immediately:

```python
sock = socket(AF_INET, SOCK_DGRAM)
sock.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
sock.setblocking(False)
sock.bind(("0.0.0.0", broadcast_port))
transport, protocol = await get_running_loop().create_datagram_endpoint(
    lambda: protocol_factory, sock=sock,
)
```

This follows the same pattern aioshelly uses for its CoAP listener socket: https://github.com/home-assistant-libs/aioshelly/blob/master/aioshelly/block_device/coap.py

## Test

Adds one focused test asserting the created listener socket has `SO_REUSEADDR` set.

Gates pass locally: `poetry run poe test` (306 passed) and `poetry run poe lint` (black, flake8, isort, mypy strict, yamllint), coverage 99.05 percent.

## Scope

Reduced per review to just the `SO_REUSEADDR` change. The datagram parse guard that was previously bundled here is split out into #890.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UDP bridge startup and socket handling for more reliable listener initialization.
  * Enabled address reuse to help prevent binding conflicts when restarting the bridge.

* **Tests**
  * Added coverage verifying that listener sockets allow address reuse.

### DIFF
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -19,7 +19,14 @@ from binascii import hexlify
 from dataclasses import dataclass
 from functools import partial
 from logging import getLogger
-from socket import AF_INET, inet_ntoa
+from socket import (
+    AF_INET,
+    SO_REUSEADDR,
+    SOCK_DGRAM,
+    SOL_SOCKET,
+    inet_ntoa,
+    socket,
+)
 from struct import error as struct_error
 from struct import pack
 from types import TracebackType
@@ -428,10 +435,17 @@ class SwitcherBridge:
             protocol_factory = UdpClientProtocol(
                 partial(_parse_device_from_datagram, self._on_device)
             )
+            # Bind with SO_REUSEADDR so a reload can rebind the port while the
+            # previous socket is still lingering in TIME_WAIT, instead of failing
+            # with "address already in use" (errno 98) until a full restart. This
+            # follows the same pattern aioshelly uses for its CoAP listener.
+            sock = socket(AF_INET, SOCK_DGRAM)
+            sock.setsockopt(SOL_SOCKET, SO_REUSEADDR, 1)
+            sock.setblocking(False)
+            sock.bind(("0.0.0.0", broadcast_port))  # nosec
             transport, protocol = await get_running_loop().create_datagram_endpoint(
                 lambda: protocol_factory,
-                local_addr=("0.0.0.0", broadcast_port),  # nosec
-                family=AF_INET,
+                sock=sock,
             )
             self._transports[broadcast_port] = transport
             logger.debug("udp bridge on port %s started", broadcast_port)
--- a/tests/test_bridge.py
+++ b/tests/test_bridge.py
@@ -101,3 +101,11 @@ async def test_malformed_datagram_is_skipped_and_logged(mock_debug):
     logged_args = mock_debug.call_args.args
     assert_that(logged_args).contains(addr)
     assert_that(logged_args).contains(data.hex())
+
+
+async def test_bridge_listener_socket_has_reuseaddr(unused_udp_broadcast_port, mock_callback):
+    port = unused_udp_broadcast_port
+    async with SwitcherBridge(mock_callback) as bridge:
+        sock = bridge._transports[port].get_extra_info("socket")
+        reuseaddr = sock.getsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR)
+        assert_that(reuseaddr).is_not_zero()


nitaybz:
Thanks for the reviews. Pushed 49e7db9 addressing the findings:

1. Raw-socket leak: `_bind_port` now closes the bound socket if `create_datagram_endpoint` fails before the transport is registered, so nothing is orphaned outside `stop()`'s view.
2. Stale delayed rebind after a rapid restart: added a lifecycle generation counter. `stop()` bumps it and cancels every pending rebind timer, and each delayed rebind captures the generation and aborts if it no longer matches, so a rebind scheduled before a `stop()` can never fire against a later `start()`, bind a duplicate, or overwrite a live transport.
3. `start()` idempotency: a call while already running logs at debug and returns instead of overwriting the live transports.
4. Parse guard narrowed: the datagram path now catches only `KeyError`, `ValueError`, `IndexError`, `struct.error`. Genuinely unexpected exceptions propagate and are not masked.
5. Bounded rebind retries: capped attempts with a capped-exponential backoff, then it gives up with a warning instead of retrying forever.
6. Aliased the asyncio `TimeoutError` import as `AsyncioTimeoutError` so it no longer shadows the builtin.
7. Tests: the fail-fast bind test now asserts `start()` raises `OSError`; the rebind test lets the loop drive `connection_lost` via `transport.abort()` rather than calling it by hand. Added coverage for the post-bind endpoint-failure socket close, the stale-rebind-after-restart guard, `start()` idempotency, the retry backoff, and the give-up cap.

Gates green locally: `poe test` (315 passed), `poe lint` (black, flake8, isort, mypy strict, yamllint), coverage 98.78 percent (fail_under 85).

Nothing intentionally skipped from the list.

nitaybz:
Hi @TomerFi, a gentle nudge on this one when you get a chance. This is the root-cause fix for the reload wedge (`OSError: [Errno 98] Address in use`) and the orphaned-socket leak that leave devices unavailable until a restart. It has been running in production across 15 Home Assistant instances for several days now with no regressions, and `poe test` + `poe lint` are green locally (315 tests, 98.8% coverage). I also folded in the CodeRabbit and Sourcery feedback in a follow-up commit.

No rush if you are busy, and glad to adjust anything you would like changed. Thanks for maintaining aioswitcher.

src/aioswitcher/bridge.py
```diff
@@ -14,12 +14,37 @@
...
+    Future,
+)
+from asyncio import TimeoutError as AsyncioTimeoutError
+from asyncio import (
+    TimerHandle,
+    ensure_future,
+    gather,
+    get_running_loop,
+    wait_for,
+)
```
TomerFi: Nit: Could you consolidate these asyncio imports into one block to clean up the top of the file?

TomerFi:
Hi @nitaybz,

Thanks for the PR. The technical approach—specifically the socket hardening with `SO_REUSEADDR`/`SO_REUSEPORT`, the awaited lifecycle management in `stop()`, and the addition of regression tests—looks solid. The production validation on 15 instances is also a helpful addition.

I don't have the hardware on hand to test this myself. Since this targets stability issues seen in the `switcher_kis` integration, I'll defer the final sign-off to @thecode and @yogevbokobza.

@thecode @yogevbokobza — can you review and let us know if this is good to merge?

src/aioswitcher/bridge.py
nitaybz: Done in 42c5e22. Consolidated into a single asyncio block, and dropped the aliased TimeoutError import in favor of import asyncio with asyncio.TimeoutError at the one use site, so the block is tidy and the builtin stays unshadowed.

nitaybz:
Thanks for the review, @TomerFi. The import nit is addressed in 42c5e22. Glad to have @thecode and @yogevbokobza take a look too, they recently merged the two related switcher_kis changes in home-assistant/core so they have the full context on the failure mode this fixes.

thecode:
@nitaybz

> SwitcherBridge can leave a broadcast port permanently deaf, and can fail to restart with OSError: [Errno 98] Address already in use. Downstream in the Home Assistant switcher_kis integration this shows up as devices going unavailable, and the integration wedging on reload until a full Home Assistant restart.

If this happens we need to fix it in Home Assistant and not try to workaround it, if the UDP socket is left open it means that other things are not cleaned up properly.

I have seen it in the past once but I currently have only a single Switcher device and was not able to reproduce it.
As I asked you on another issue, if you are trying to fix an issue please provide logs so we understand what is the issue and how to fix it.

There are many integrations uses UDP and none of them implemented a complex logic to rebind to a port, I think this is not the right direction and I am a little bit afraid that the logic will introduce other problems, especially using long timeout of 30 seconds which is longer than timeout used in Home Assistant and will end up with more mess.

`SO_REUSEADDR` is good anyhow (not `SO_REUSEPORT`, we use broadcasts and don't send anything so we don't need it) and would work even if the socket is not closed on previous attempt, see an example in `aioshelly`:

https://github.com/home-assistant-libs/aioshelly/blob/42ccefbe137145954257cda0351527e23c0b9ecd/aioshelly/block_device/coap.py#L117-L125

This simple implementation above exists for more than 6 years any problems, you can easily test it buy running two processes that listen to the same port in parallel and they will receive data.

But as I started, I would like first to see a debug log of a scenario that creates this error and fix it before we try to workaround it.

src/aioswitcher/bridge.py
```diff
@@ -460,18 +649,44 @@ def is_running(self) -> bool:
...
+        Guarded so a malformed or unknown-model datagram is dropped instead of
+        raising into the event loop, which would otherwise interrupt reception
+        for every other device on the same socket. Only the expected parse
+        errors are swallowed, a genuinely unexpected exception still propagates
+        so it is not masked.
+        """
+        try:
+            self._on_datagram(data)
+        except _PARSE_EXCEPTIONS:
+            logger.debug("dropping an unparseable switcher datagram", exc_info=True)
```
thecode: @nitaybz This seems to be a good way to guard for errors, same as I wrote you on Home Assistant, if you can split PRs for changes that does not have to be together it would be better because it is easier to review and approve.

nitaybz: Done, split out into #890 which carries only this parse guard. This PR is now just the SO_REUSEADDR change.

nitaybz:
Thanks @thecode, that makes sense. I have reduced this PR to just the minimal SO_REUSEADDR fix on the broadcast listener socket, following the aioshelly CoAP pattern you linked, and dropped the SO_REUSEPORT and rebind logic. The datagram parse guard is now split out into #890 so each can be reviewed on its own.

On the logs point: SO_REUSEADDR is the minimal, endorsed fix here and stands on its own, it just lets a reload rebind the port that is still lingering in TIME_WAIT instead of hitting errno 98. Happy to capture a repro debug log if it would help with anything further.

thecode:
> Thanks @thecode, that makes sense. I have reduced this PR to just the minimal SO_REUSEADDR fix on the broadcast listener socket, following the aioshelly CoAP pattern you linked, and dropped the SO_REUSEPORT and rebind logic. The datagram parse guard is now split out into #890 so each can be reviewed on its own.

Thanks for the fast response and addressing the feedback, this change is smaller and easier to review, I will test it in the following days. I suggest we target this change to the next Home Assistant release (although it is possible to push this as a bugfix) which is 4 weeks from now so we have time to create new `aioswitcher` release and test it.

> On the logs point: SO_REUSEADDR is the minimal, endorsed fix here and stands on its own, it just lets a reload rebind the port that is still lingering in TIME_WAIT instead of hitting errno 98. Happy to capture a repro debug log if it would help with anything further.

I would love to fix this, the best is to first fix it before we bump `aioswitcher`. 
Please feel free to contact me on discord if you have logs or reproduce it, I am a little bit overloaded with tasks across few integrations but if you can help reproduce we can easily fix it.

Thanks for your contributions to `aioswitcher` & `switcher_kis` 👍

nitaybz:
Following up on the repro you asked for. Instead of pasting redacted customer logs, here is a minimal self-contained reproduction of the exact behavior on the Switcher broadcast port. Nothing confidential in it, and you can run it yourself.

What we saw in production: on a config-entry reload (and the occasional double-setup, home-assistant/core#140961), the new `SwitcherBridge.start()` binds the broadcast port before the previous listener's socket has been released, and the bind fails with `OSError: [Errno 98] Address already in use`. That cascades into `Config entry ... for switcher_kis.<platform> has already been setup` and the integration stays wedged until a full Home Assistant restart. `start()` binds via `create_datagram_endpoint(local_addr=...)`, and since Python 3.8 asyncio no longer sets `SO_REUSEADDR` implicitly for datagram endpoints (it was dropped for UDP for security reasons), so nothing sets it and the reload rebind collides. Same reason aioshelly sets it on its CoAP listener.

Minimal reproduction (Python 3.12, Linux, matching the HA container):

```python
import socket

PORT = 20002  # Switcher UDP broadcast port

def bind(flags):
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    for f in flags:
        s.setsockopt(socket.SOL_SOCKET, f, 1)
    s.bind(("0.0.0.0", PORT))
    return s

def show(label, flags):
    a = bind(flags)
    try:
        b = bind(flags)
        print(f"{label:<14} second listener binds OK")
        b.close()
    except OSError as e:
        print(f"{label:<14} second listener FAILS -> [Errno {e.errno}] {e.strerror}")
    a.close()

show("no options", [])
show("SO_REUSEADDR", [socket.SO_REUSEADDR])
```

Output:

```
no options     second listener FAILS -> [Errno 98] Address already in use
SO_REUSEADDR   second listener binds OK
```

The two active listeners model the reload window where the old and new bridge sockets briefly coexist. Without `SO_REUSEADDR` the second bind fails, which is the wedge. With it (this PR) the reloading bridge binds cleanly. Happy to dig up anything else that would help.

thecode (dismissed):

thecode:
@TomerFi @nitaybz I tested the changes in this PR and they are ready to be merged, but merging is blocked due to commit signatures not verified

TomerFi:
@nitaybz @thecode After merging #890, this one now has conflicts.

nitaybz:
Rebased onto the current `dev` to clear the merge conflict — this branch was
conflicting with #890, which touched the same import block and the end of
`tests/test_bridge.py`.

The conflict was textual only; both sides are kept as-is:

- `src/aioswitcher/bridge.py` — the `socket` import is merged into the single
  parenthesised form that includes `SO_REUSEADDR`/`SOCK_DGRAM`/`SOL_SOCKET`
  alongside `struct_error` from #890. No logic change.
- `tests/test_bridge.py` — `test_malformed_datagram_is_skipped_and_logged`
  (#890) and `test_bridge_listener_socket_has_reuseaddr` (this PR) both kept.

The diff is unchanged from what was approved (25 insertions, 3 deletions across
the same two files). Locally: `tests/test_bridge.py` 6/6 pass, full suite
306 passed, `flake8`/`isort`/`black`/`mypy` clean.

No action needed on your side beyond a re-approval — the force-push dismissed
the previous one. Thanks!

thecode (approved):

TomerFi:
@nitaybz @thecode - Thank you! This was included in [6.1.3](https://github.com/TomerFi/aioswitcher/releases#release-6.1.3).

PR #890, opened 2026-07-05
fix: guard datagram parsing against malformed packets

nitaybz:
## Problem

`UdpClientProtocol.datagram_received` parses each incoming datagram inline. A malformed or unknown-model packet (for example a device model byte that maps to no known `DeviceType`, which raises `KeyError` in `get_device_type`, or a truncated packet) propagates the exception into the event loop, which interrupts reception for every other device on that socket.

## The fix

Catch the expected parse errors in `datagram_received`, log the offending packet at debug level, and skip it:

```python
try:
    self._on_datagram(data)
except (KeyError, ValueError, IndexError, struct.error):
    logger.debug("dropping an unparseable switcher datagram", exc_info=True)
```

Only the expected parse exceptions are caught, so a genuinely unexpected error still propagates and is not masked.

## Test

Adds a focused test that a datagram whose parse raises (an unknown model mapping to `KeyError`) is dropped rather than raised.

Gates pass locally: `poetry run poe test` (306 passed) and `poetry run poe lint` (black, flake8, isort, mypy strict, yamllint), coverage 99.05 percent.

## Context

Split out from #887 at @thecode's request, so the SO_REUSEADDR reload fix and this parse guard can be reviewed independently. #887 now carries only the SO_REUSEADDR change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Malformed or unexpected UDP packets no longer cause the app to crash or interrupt background processing.
  * Problem packets are now safely skipped and logged for debugging, while normal traffic continues to be handled as expected.
  * Added coverage to ensure invalid datagrams are ignored and recorded properly.

### DIFF
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -20,6 +20,7 @@ from dataclasses import dataclass
 from functools import partial
 from logging import getLogger
 from socket import AF_INET, inet_ntoa
+from struct import error as struct_error
 from struct import pack
 from types import TracebackType
 from typing import Any, Callable, Dict, Optional, Tuple, Type, final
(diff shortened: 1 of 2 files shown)


thecode (dismissed):
Thanks!

thecode:
Would be nice to address https://github.com/TomerFi/aioswitcher/pull/890#pullrequestreview-4631972319 since it will help understanding in future logs if the error is from a real Switcher device or what is the offending device that makes it

I was helping capturing a malformed packet on another project and it took years (literally years) until someone successfully captured the malformed packet since the lib was not logging who is the sender, what is the packet and the error occurred one a month (which is hard to leave a tcpdump running for such a long time), a log of the packet and the sender could shorten this to days instead of years.

nitaybz:
Thanks for the approval and the context, @thecode. Good point on traceability. I enhanced the parse-guard debug log in ae75f40 so it now includes the sender address, the raw packet as hex, and the error:

    logger.debug("failed to parse datagram from %s: %s (data=%s)", addr, error, data.hex())

So a future malformed packet can be traced back to who sent it and what the bytes were, instead of being silently dropped.

thecode (approved):

thecode:
@TomerFi @nitaybz I tested the changes in this PR and they are ready to be merged, but merging is blocked due to commit signatures not verified

TomerFi:
@nitaybz @thecode - Thank you! This was included in [6.1.3](https://github.com/TomerFi/aioswitcher/releases#release-6.1.3).

PR #924, opened 2026-08-22
fix: allow durations of 24 hours or more in seconds_to_iso_time

tomer-w:
## Problem

`seconds_to_iso_time` builds its result via `datetime.time(hour=hours, ...)`, which raises `ValueError: hour must be in 0..23` whenever the computed hour reaches 24 or more.

This function is used to format *durations* (elapsed seconds), not times-of-day, e.g. `SwitcherStateResponse.time_on` and `.time_left`. As a result, `SwitcherApi.get_state()` fails for any power plug (or other Type 1 device) that has been powered on for 24 hours or longer, surfacing as a misleading:

```
RuntimeError: get state request was not successful
```

## Root cause confirmed against a real device

Using a read-only diagnostic script (login + get-state only, no control packets), a real `SwitcherPowerPlug` on my network returned a valid state response with `time_on = 702294` seconds (195h 04m 54s). Parsing that value raised the `ValueError` above, which `get_state()` converts into the generic `RuntimeError`.

## Fix

- Format the duration manually (`f"{hours:02}:{minutes:02}:{seconds:02}"`) instead of constructing a `datetime.time`, removing the 0-23 hour cap while keeping the existing `HH:MM:SS`-shaped return value and negative-input validation.
- Updated the docstring to clarify this is a duration formatter, not a time-of-day formatter.
- Added a regression test covering a duration beyond 24 hours (`702294` -> `"195:04:54"`), and adjusted the existing negative-value test to match the new error message.

## Testing

- `uv run pytest` - full suite passes (one pre-existing, unrelated clock-boundary flake in `test_schedule_tools.py` that passes in isolation and is unaffected by this change)
- `uv run ruff check` / `uv run ruff format --check` - clean on changed files
- `uv run ty check` - clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Duration formatting now correctly supports values of 24 hours or longer.
  - Negative durations are rejected with a clear validation error.
  - Time output remains consistently formatted as `HH:MM:SS` across device, heater, and automatic shutdown timers.
  - Remaining-time and scheduled shutdown displays now use consistent duration formatting.

- **Tests**
  - Added coverage for multi-day durations, negative-input validation, and timer-related formatting scenarios.

### DIFF
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -30,7 +30,7 @@ from ..device import DeviceType
 logger = getLogger(__name__)
 
 
-def seconds_to_iso_time(all_seconds: int) -> str:
+def seconds_to_duration_string(all_seconds: int) -> str:
     """Convert seconds to an HH:MM:SS duration string.
 
     Args:
--- a/src/aioswitcher/device/tools.py
+++ b/src/aioswitcher/device/tools.py
@@ -31,20 +31,25 @@ logger = getLogger(__name__)
 
 
 def seconds_to_iso_time(all_seconds: int) -> str:
-    """Convert seconds to iso time.
+    """Convert seconds to an HH:MM:SS duration string.
 
     Args:
         all_seconds: the total number of seconds to convert.
 
     Return:
-        A string representing the converted iso time in %H:%M:%S format.
-        e.g. "02:24:37".
+        A string representing the converted duration in %H:%M:%S format.
+        e.g. "02:24:37". Unlike a time-of-day, the hours component is not
+        capped at 23, so durations of a day or longer are represented
+        correctly, e.g. "195:04:54".
 
     """
+    if all_seconds < 0:
+        raise ValueError("seconds cannot be negative")
+
     minutes, seconds = divmod(int(all_seconds), 60)
     hours, minutes = divmod(minutes, 60)
 
-    return datetime.time(hour=hours, minute=minutes, second=seconds).isoformat()
+    return f"{hours:02}:{minutes:02}:{seconds:02}"
 
 
 def sign_packet_with_crc_key(hex_packet: str) -> str:
--- a/src/aioswitcher/api/messages.py
+++ b/src/aioswitcher/api/messages.py
@@ -30,7 +30,7 @@ from ..device import (
 from ..device.tools import (
     get_light_discovery_packet_index,
     get_shutter_discovery_packet_index,
-    seconds_to_iso_time,
+    seconds_to_duration_string,
     watts_to_amps,
 )
 from ..schedule.parser import SwitcherSchedule, get_schedules
@@ -62,7 +62,7 @@ class StateMessageParser:
             + hex_time_left[0:2],
             16,
         )
-        return seconds_to_iso_time(time_left_seconds)
+        return seconds_to_duration_string(time_left_seconds)
 
     def get_time_on(self) -> str:
         """Return how long the device has been on."""
@@ -71,7 +71,7 @@ class StateMessageParser:
             hex_time_on[6:8] + hex_time_on[4:6] + hex_time_on[2:4] + hex_time_on[0:2],
             16,
         )
-        return seconds_to_iso_time(time_on_seconds)
+        return seconds_to_duration_string(time_on_seconds)
 
     def get_auto_shutdown(self) -> str:
         """Return the value of the auto shutdown configuration."""
@@ -83,7 +83,7 @@ class StateMessageParser:
             + hex_auto_off[0:2],
             16,
         )
-        return seconds_to_iso_time(auto_off_seconds)
+        return seconds_to_duration_string(auto_off_seconds)
 
     def get_state(self) -> DeviceState:
         """Return the current device state."""
@@ -191,7 +191,7 @@ class StateMessageParser:
             + hex_time_left[0:2],
             16,
         )
-        return seconds_to_iso_time(time_left_seconds)
+        return seconds_to_duration_string(time_left_seconds)
 
     def get_heater_time_on(self) -> str:
         """Return how long the heater device has been on."""
@@ -200,7 +200,7 @@ class StateMessageParser:
             hex_time_on[6:8] + hex_time_on[4:6] + hex_time_on[2:4] + hex_time_on[0:2],
             16,
         )
-        return seconds_to_iso_time(time_on_seconds)
+        return seconds_to_duration_string(time_on_seconds)
 
     def get_heater_auto_shutdown(self) -> str:
         """Return the value of the heater auto shutdown configuration."""
@@ -212,7 +212,7 @@ class StateMessageParser:
             + hex_auto_off[0:2],
             16,
         )
-        return seconds_to_iso_time(auto_off_seconds)
+        return seconds_to_duration_string(auto_off_seconds)
 
     def get_heater_state(self) -> DeviceState:
         """Return the current heater device state."""
--- a/src/aioswitcher/bridge.py
+++ b/src/aioswitcher/bridge.py
@@ -55,7 +55,7 @@ from .device import (
 from .device.tools import (
     get_light_discovery_packet_index,
     get_shutter_discovery_packet_index,
-    seconds_to_iso_time,
+    seconds_to_duration_string,
     watts_to_amps,
 )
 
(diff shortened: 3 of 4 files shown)


thecode (dismissed):

tomer-w:
Why do I need to sign the commits? Who care who I am? If the commits are good, why isn't it enough?

TomerFi:
> Why do I need to sign the commits? Who care who I am? If the commits are good, why isn't it enough?

Commits must be verified.

src/aioswitcher/device/tools.py
```diff
@@ -31,20 +31,25 @@
 
 
 def seconds_to_iso_time(all_seconds: int) -> str:
-    """Convert seconds to iso time.
+    """Convert seconds to an HH:MM:SS duration string.
```
TomerFi, line 34: The function _second_to_iso_time_ implies an ISO 8601 clock time (max 24 hours). We should audit its callers: if it's strictly used for durations, we need to rename it to reflect that. If it was meant for clock times, we should confirm why and verify if that usage is still valid before making any changes.

TomerFi (changes requested):

src/aioswitcher/device/tools.py
tomer-w: I went over all callers, and it seems they are all talking about duration. I see 3 families of usage:
1. How long the device is on -> Can be any duration. This is where I discovered the failure.
2. What is the auto-shutdown duration -> Also duration but theoretically can be capped by the device manufacturer. That cap doesn't have to be 24 hours.
3. Remaining time till shutdown -> Duration as well but capped by the auto-shutdown duration.
I don't see any clock time usage so based on that I will prepare a commit which rename the function.

tomer-w:
@TomerFi, I renamed the function to ```seconds_to_duration_string``` based on your comment. As this is public function theoretically it can be used by other projects, although I will be surprised if it does. There was an option to keep alias to the previous name but based on your feedback I assumed this will be confusing, so I opted not to do it.

tomer-w:
> > Why do I need to sign the commits? Who care who I am? If the commits are good, why isn't it enough?
> 
> Commits must be verified.

I thought tests verify commit.

src/aioswitcher/device/tools.py
```diff
@@ -30,21 +30,26 @@
 logger = getLogger(__name__)
 
 
-def seconds_to_iso_time(all_seconds: int) -> str:
-    """Convert seconds to iso time.
+def seconds_to_duration_string(all_seconds: int) -> str:
```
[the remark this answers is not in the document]
TomerFi: @coderabbitai
Niether switcher_webapi nor switcher_kis use this fucntion directly. Perhaps we can prefix it with an underscore to clarify its internal intent. But this is not the only internal function we have here, so this can go in another PR.

TomerFi:
> @TomerFi, I renamed the function to `seconds_to_duration_string` based on your comment. As this is public function theoretically it can be used by other projects, although I will be surprised if it does. There was an option to keep alias to the previous name but based on your feedback I assumed this will be confusing, so I opted not to do it.

@tomer-w Thank you for the quick turn around. Yes, you were right about the usage and for skipping the alias. It's an internal function. We should probably prefix it with an underscore. But that's an idea for a different PR. Regardless, this commit changes the return context enough to be considered a breaking change anyway. So I'm ok with the name change.

@thecode From Home Assistant prespective, are we ok with this change for the scenarios described [here](https://github.com/TomerFi/aioswitcher/pull/924#discussion_r3837771940): 
```
1. How long the device is on -> Can be any duration. This is where I discovered the failure.
2. What is the auto-shutdown duration -> Also duration but theoretically can be capped by the device manufacturer. That cap doesn't have to be 24 hours.
3. Remaining time till shutdown -> Duration as well but capped by the auto-shutdown duration.
```

TomerFi:
> I thought tests verify commit.

CI checks and contributor reviews validate the commit's code, whereas digital signing confirms who wrote it. See GitHub documentation [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).

tomer-w:
> > @TomerFi, I renamed the function to `seconds_to_duration_string` based on your comment. As this is public function theoretically it can be used by other projects, although I will be surprised if it does. There was an option to keep alias to the previous name but based on your feedback I assumed this will be confusing, so I opted not to do it.
> 
> @tomer-w Thank you for the quick turn around. Yes, you were right about the usage and for skipping the alias. It's an internal function. We should probably prefix it with an underscore. But that's an idea for a different PR. Regardless, this commit changes the return context enough to be considered a breaking change anyway. So I'm ok with the name change.
> 
> @thecode From Home Assistant prespective, are we ok with this change for the scenarios described [here](https://github.com/TomerFi/aioswitcher/pull/924#discussion_r3837771940):
> 
> ```
> 1. How long the device is on -> Can be any duration. This is where I discovered the failure.
> 2. What is the auto-shutdown duration -> Also duration but theoretically can be capped by the device manufacturer. That cap doesn't have to be 24 hours.
> 3. Remaining time till shutdown -> Duration as well but capped by the auto-shutdown duration.
> ```

If you want me to make it private (which I think the right thing to do), I have no problem with another small commit.

With regards for HA, I dont think this should create any problem. As those fields are string, HA will not care if those strings represent something bigeer than 24 hours. It has no understanding it is duration related to begin with.

tomer-w:
> > I thought tests verify commit.
> 
> CI checks and contributor reviews validate the commit's code, whereas digital signing confirms who wrote it. See GitHub documentation [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification).

In OSS projects when you dont really know me, the fact that I will generate some key pair will not help you in any way. All you know I'm @tomer-w whatever it means. HA and most of the large OSS projects dont require that verification as it is really meaningless. You approve commits by the quality of the code and not by the fact you know who I am. I know it is somehow used with very deep cryptographic project when the code is so complex than trusting the author is big part of allowing some commits. This is not the case here.

thecode:
> 1. How long the device is on -> Can be any duration. This is where I discovered the failure.

This is a string in Home Assistant so there is no limit, I have never seen a value more than 24 Hours, but we already know that Switcher has different behavior on their Plugs

> 2. What is the auto-shutdown duration -> Also duration but theoretically can be capped by the device manufacturer. That cap doesn't have to be 24 hours.

I am not aware of a value more than 24 hours, please note that "capped by the device manufacturer" is something we need to handle, meaning we have to test what is the allowed value (send it and see if it is working) and limit it, Switcher does not respond with a value that can be used to know if a value is accepted or not. This is again a string in Home Assistant so no validation.

> 3. Remaining time till shutdown -> Duration as well but capped by the auto-shutdown duration.

Note that here, when using the turn on with timer service there is a validation of the value in Home Assistant:
YAML: https://github.com/home-assistant/core/blob/661fddbe0fa38d1aee11d4d81c58bb0e3ef0ba50/homeassistant/components/switcher_kis/services.yaml#L25-L26
Docs: https://www.home-assistant.io/actions/switcher_kis.turn_on_with_timer/#options-in-the-ui

Both have to be updated to allow a larger value.

tomer-w:
I don't suggest allowing to send values larger than 24 hours. The change I made is just allowing formatting of the string not to fail when the device itself send the value. If the device send it, than it have to be legal.

tomer-w:
@TomerFi , anything which is blocking the merge?

TomerFi:
> TomerFi , anything which is blocking the merge?

@tomer-w Thank you very much for your contribution. I prefer to wait for @thecode 's approval for this one.

Also, @thecode , you maintain multiple other projects, what's the status on the verified commits requirement over there? Is this no longer considered a best practice in OSS projects?

thecode:
> @tomer-w Thank you very much for your contribution. I prefer to wait for @thecode 's approval for this one.

I was sure I already did, now I noticed that I didn't noticed the last commit, anyhow it is good to go.

> Also, @thecode , you maintain multiple other projects, what's the status on the verified commits requirement over there? Is this no longer considered a best practice in OSS projects?

There is no requirement for signed commits in any of the projects I am working on

thecode (approved):

TomerFi:
https://github.com/TomerFi/aioswitcher/releases/tag/6.3.0

tomer-w:
Thanks guys!

Issue #929, opened 2026-09-04
Update IR database for Switcher Breeze (ELEC7024)


oranav:
The IR database is missing some IR codes, like ELEC7024. I would like to update it.

Not sure how we obtained the database in the first place; if we got it through an official channel - can we please request an updated one?
If it has been reverse engineered from the app itself (or obtained through the Switcher cloud API), I can take on that from scratch, or - if we have some scripts - that would be very helpful.

I do see that `https://switcher.co.il/misc/irGet/getIR.php` is no longer available.

oranav:
In the meantime, I extracted the codes from my app's sqlite database, and raised a PR here: #930 
Would absolutely love to get it merged, so I can control my AC using Home Asssitant ;)

PR #904, opened 2026-07-29
chore: update ir database

YogevBokobza:
## Description

IR database from Switcher updated.
Including remote IDS: SPE20711 and HAER7002

## Checklist

- [x] I have followed this repository's contributing guidelines.
- [x] I will adhere to the project's code of conduct.

PR #930, opened 2026-09-04
feat: add ELEC7024 Breeze remote

oranav:
## Summary

- add the ELEC7024 remote to the bundled Breeze IR-set database
- include all 249 extracted commands with the shared NECX protocol parameters
- preserve its toggle-style OnOffType behavior

## Validation

- validated the JSON and all 249 command keys and 18-byte hex payloads
- ran the installed pre-commit hooks successfully
- ran tests/test_api_tcp_client.py: 58 passed
- tested against a local Switcher Breeze using ELEC7024:
  - turned the AC off and confirmed the device reported OFF
  - turned it on in cool mode at 23 degrees Celsius with automatic fan and confirmed all settings in the device response

## Summary by CodeRabbit

- **New Features**
  - Added support for the ELEC7024 device.
  - Added infrared commands for power, temperature, air direction, and fan-speed controls.


thecode:
I remembered that I have seen this remote already, seems like it was added at https://github.com/TomerFi/aioswitcher/pull/904/changes/c8a1558167050b47bb9e59420ae952a6568ac83e https://github.com/TomerFi/aioswitcher/pull/904 and later removed when we found duplicated remotes with different data and @YogevBokobza asked Switcher for a fixed file.

I would prefer that Switcher will be the source of the data, if we accept this PR, on the next file update your contribution will be lost since it is impossible to merge such a large file and we just replace the file.

@YogevBokobza @oranav I suggest you contact Switcher and let them explain why it is missing or send us an updated remote file.

oranav:
Thanks @thecode.

> I remembered that I have seen this remote already, seems like it was added at [c8a1558](https://github.com/TomerFi/aioswitcher/commit/c8a1558167050b47bb9e59420ae952a6568ac83e) #904 and later removed when we found duplicated remotes with different data and @YogevBokobza asked Switcher for a fixed file.

Are you sure it was added on the commit you mentioned? I don't see any mention of ELEC7024 there, and grepping yields no results. Are you maybe confused with other codes like ELEC7021?

```shell
$ git grep ELEC7024 $(git rev-list HEAD) || echo Nothing
Nothing
```

I doubt there are other remotes with the exact same code. When I initially set up the Switcher app, it only provided 3 options - and only one of them (ELEC7024) worked with my AC properly.

> I would prefer that Switcher will be the source of the data, if we accept this PR, on the next file update your contribution will be lost since it is impossible to merge such a large file and we just replace the file.

Agree. I mentioned this on #929. That's the better solution (rather than reverse engineering it from the app itself).

> @YogevBokobza @oranav I suggest you contact Switcher and let them explain why it is missing or send us an updated remote file.

How should we contact them? Just reach out to office@switcher.co.il? Or do we have a technical POC there?

oranav:
@thecode @TomerFi @YogevBokobza Gentle ping on this. Would really like to get my AC working with HA 😇
Shanah tovah!

TomerFi:
> @thecode @TomerFi @YogevBokobza Gentle ping on this. Would really like to get my AC working with HA 😇 Shanah tovah!

Shanna Tova @oranav and thank you for this PR!
I'll be happy to help with merging and releasing once @thecode and @YogevBokobza approve this.

thecode:
@TomerFi  from my side nothing has changed from what I wrote at https://github.com/TomerFi/aioswitcher/pull/930#issuecomment-5547753510

1. Switcher has already supplied us a file with this AC, but it had errors and when they fixed it, they removed this AC, I don't know why. This is not the only AC they removed (see https://github.com/TomerFi/aioswitcher/commit/c8a1558167050b47bb9e59420ae952a6568ac83e for a larger file that has this AC)
2. We agreed with Switcher that they will supply us new files when ACs are updated, if we manually merge data into this file we will not be able to update this file in the future just by replacing the file.

Someone need to contact Switcher and ask for an updated/fixed file. I don't have the capacity to do it or an updated contact. @YogevBokobza was the last one in contact with them.

oranav:
@thecode Sorry, you're right. This AC was indeed in c8a1558167050b47bb9e59420ae952a6568ac83e, which never had been merged into the dev branch (the PR had been squash merged).

I'll be happy to take it with Switcher; I reached out to office@switcher.co.il but got no response so far.

oranav:
@thecode @TomerFi Since I don't get any response from office@switcher.co.il, and this PR has been open for 2 weeks now - WDYT about getting this merged, and later when we get an updated file from Switcher - we can override the file with their version? I don't think we should care too much about diffing. We can just use their version - as long as it includes all ACs.

Holding this off means we're missing support for some common ACs in Israel (including mine 😉).

Chatimah tovah!

TomerFi:
@oranav First, thank you for your work on this PR and for your continued efforts to get a response from the vendor. Because this change could impact Home Assistant—the primary client of this library—it cannot be merged without explicit approval from @thecode or @YogevBokobza. Since Switcher devices aren't available in my region, I have no way to test this firsthand and want to avoid rubber-stamping it.

Chatima tovah!

oranav:
@TomerFi Absolutely understood. As a matter of fact, the (kind of selfish) goal of this PR is to enable Home Assistant to support my own AC ;)

I don't see a big risk here though: this PR doesn't change any existing IR code - but rather adds a new one (which doesn't exist in the database) - and I can assure you I've tested it on my AC 😊
The biggest risk IMO is that one of the codes here is wrong, and in that case - ELEC7024 will only work partially (instead of not at all, which is the current state).

@thecode I would love your approval here. As I said before: if we can get an updated IR database from Switcher soon - let's do it and drop this PR; otherwise, I suggest that we merge this for the short term - until we get an updated database from them.

